PR #1, opened 2016-12-02
First version of Open API spec for v4 API

marians:
A first skeleton for the spec


oponder:
LGTM to get some kind of 'base' of code into this repo.

Missing some response details and stuff in the examples, but lets build on this instead of waiting to get it perfect

PR #2, opened 2017-01-19
added table with more response codes

xh3b4sd:
This PR extends the list of available response codes within the API. Here I took the freedom to suggest another format of documentation. I would say it is good enough to have a table with the actual code and a broad description of what it means. The details should be covered in the corresponding API documentation. So the docs to the response codes is only the about the actual code and its purpose. IMO we do not need to list all routes in the documentation about response codes.


xh3b4sd:
@marians I applied your feedback. Merging.

PR #12, opened 2017-04-11
Adds /v4/ping/ spec

JosephSalisbury:
Towards https://github.com/giantswarm/api/issues/355
Towards https://github.com/giantswarm/api/issues/280

As an implementation note, this route will ping `userd`, `tokend`, `companyd`, and `cluster-service`'s `/healthz` routes, with a timeout. We'll need to add those endpoints (see https://github.com/giantswarm/giantswarm/issues/1104 and https://github.com/giantswarm/giantswarm/issues/1118).

We can also reuse `/v4/ping/` internally as `/healthz`

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -146,3 +146,12 @@ definitions:
       client_certificate_data:
         type: string
         description: PEM-encoded certificate
+
+  V4PingResponse:
+    type: object
+    required:
+      - message
+    properties:
+      message:
+        type: string
+        description: Ping status
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -146,12 +146,3 @@ definitions:
       client_certificate_data:
         type: string
         description: PEM-encoded certificate
-
-  V4PingResponse:
-    type: object
-    required:
-      - message
-    properties:
-      message:
-        type: string
-        description: Ping status
--- a/spec.yaml
+++ b/spec.yaml
@@ -259,3 +259,30 @@ paths:
                 "id": "02:cc:da:f9:fb:ce:c3:e5:e1:f6:27:d8:43:48:0d:37:4a:ee:b9:67",
                 "ttl_hours": 8640
               }
+
+  /v4/ping/:
+    get:
+      operationId: ping
+      tags:
+        - ping
+      summary: Ping the API connection
+      description: This operation tests the connection to the API, and performs an internal health check.
+      responses:
+        "200":
+          description: OK
+          schema:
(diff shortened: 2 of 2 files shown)


JosephSalisbury:
See http://apispec.g8s.fra-1.giantswarm.io/v4-ping/#tag/ping

oponder:
For bonus points: you could explain in the description what the 500 response means... "Will return 500 when the internal health check fails"

LGTM

JosephSalisbury:
@marians mind having another look?

marians:
@JosephSalisbury BTW, don't hit merge just yet, please, as `master` should represent what's really implemented.

xh3b4sd:
As we talked about this in the API sync I want to drop my 2 cents. I see no reason to have health and ping routes. I would like to have one concept to rule them all if that makes sense. So maybe we can simply have `/healthz/` (I don't like that fancy z bullshit) or `/v4/healthz/`.

JosephSalisbury:
@xh3b4sd do you mean `/health/`?

oponder:
I don't think we should try to make everything that is very similar into one concept. I think we try sometimes to DRY things too much. 

> Duplication is far cheaper than the wrong abstraction

And in this case its not even duplication of code; its duplication in the sense of having two endpoints that will probably call the same bit of code for determining its reply.

Users want to know the state of the system and k8s wants to know the state of the system.
These are two different users with possibly different demands in the future. Even though right now we use the same logic for both cases, it does not mean that will be the case forever.

The way I see the `z` at the end of `healthz` is that it is a convention (still in its infancy) for endpoints that are meant to be consumed by monitoring / alerting / and automated recovery services: https://github.com/search?p=2&q=healthz&type=Repositories&utf8=%E2%9C%93

So i would vote for using the fancy `z` endpoint.

If we argue "YAGNI" (you aren't going to need it) then I would agree, and say:

1. Implement `/v4/ping` and tell our manifests that this is the health endpoint.
2. Wait until this no longer serves our needs (i.e., our system health check needs something that we do not want to expose to our users)
3. Implement `/healthz` and edit manifests then to say that this is the health endpoint, and keep `/v4/ping`

xh3b4sd:
I really don't understand that, but I do not need to understand everything. When you think it is right, go ahead on your own behalf. Disagree and commit.

JosephSalisbury:
My two cents would be `/v4/ping` as the API spec, and `/healthz` as the route we have for the manifests

JosephSalisbury:
This is so fucking old now good bye

PR #13, opened 2017-04-19
Add more details to possible cluster error responses

oponder:
This specifies what kind of errors our Cluster endpoints will return, as this is not yet specified.

In a nutshell: I think the API should return 401 when the user tries to create a cluster but the organization does not exist or the user is not a member of the organization. (As is already implemented)

However, for the GET, DELETE, and PATCH requests I would expect the API to return a 404, even when the cluster actually exists, but the user simply can't act on it because he is not a member of the organization.

The reasoning behind this difference is that in the 'CREATE' case, we are not acting on something that exists yet, so it doesn't make sense to return a 404. 

In the case of GET, DELETE, and PATCH, the user is providing a `cluster_id` and wants to act on something that he thinks is there. So when he can't act on it because it doesn't exist, or because he shouldn't have access to it, we should return a 404. We return a 404 instead of a 401 in the 'he shouldn't have access it' case because we want to avoid revealing any information we don't need to. If we we're to return a 401, we'd reveal that the cluster ID is in use.

------------------------------

**POST** `/v4/clusters/` will return `401` when the user tries to create a cluster for an organization that he is not a member of. (Already implemented)

**POST** `/v4/clusters/` will return `401` when the user tries to create a cluster for an organization that does not exist. (Already implemented)

**GET** `/v4/clusters/{cluster-id}` will return `404` when the cluster doesn't exist

**GET** `/v4/clusters/{cluster-id}` will return `404` when the cluster exists but the user is not a member of the organization the cluster belongs to.

**DELETE** `/v4/clusters/{cluster-id}` will return `404` when the cluster doesn't exist

**DELETE** `/v4/clusters/{cluster-id}` will return `404` when the cluster exists but the user is not a 
member of the organization the cluster belongs to.

**PATCH** `/v4/clusters/{cluster-id}` will return `404` when the cluster doesn't exist

**PATCH** `/v4/clusters/{cluster-id}` will return `404` when the cluster exists but the user is not a 
member of the organization the cluster belongs to.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -3,9 +3,6 @@ definitions:
   # Generic response model
   V4GenericResponse:
     type: object
-    required:
-      - code
-      - message
     properties:
       message:
         type: string
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -9,8 +9,10 @@ definitions:
     properties:
       message:
         type: string
+        description: "A human readable message"
       code:
         type: string
+        description: "One of the response codes from: https://github.com/giantswarm/api-spec/blob/master/details/RESPONSE_CODES.md"
 
   # Request to create a new cluster
   V4AddClusterRequest:
--- a/spec.yaml
+++ b/spec.yaml
@@ -99,6 +99,10 @@ paths:
                 "code": "PERMISSION_DENIED",
                 "message": "The requested resource cannot be accessed using the provided permission rights. Please make sure you are a member of the organization you set in the 'owner' field."
               }
+        default:
+          description: error
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/clusters/{cluster_id}/:
     get:
@@ -161,6 +165,10 @@ paths:
                 "code": "RESOURCE_NOT_FOUND",
                 "message": "The cluster with ID 'wqtlq' could not be found, or perhaps you do not have access to it. Please make sure the cluster ID is correct, and that you are a member of the organization that it belongs to."
               }
+        default:
+          description: error
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
     patch:
       operationId: modifyCluster
       tags:
@@ -189,6 +197,10 @@ paths:
                 "code": "RESOURCE_NOT_FOUND",
                 "message": "The cluster with ID 'wqtlq' could not be found, or perhaps you do not have access to it. Please make sure the cluster ID is correct, and that you are a member of the organization that it belongs to."
               }
+        default:
+          description: error
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
     delete:
       operationId: deleteCluster
       tags:
@@ -223,6 +235,10 @@ paths:
                 "code": "RESOURCE_NOT_FOUND",
                 "message": "The cluster with ID 'wqtlq' could not be found, or perhaps you do not have access to it. Please make sure the cluster ID is correct, and that you are a member of the organization that it belongs to."
               }
+        default:
+          description: error
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/clusters/{cluster_id}/key-pairs/:
     get:
@@ -242,7 +258,7 @@ paths:
           schema:
             $ref: "definitions.yaml#/definitions/V4GetKeyPairsResponse"
         default:
-          description: Error
+          description: error
           schema:
             $ref: "definitions.yaml#/definitions/V4GenericResponse"
     post:
--- a/spec.yaml
+++ b/spec.yaml
@@ -159,7 +159,7 @@ paths:
             application/json:
               {
                 "code": "RESOURCE_NOT_FOUND",
-                "message": "The cluster could not be found."
+                "message": "The cluster with ID 'wqtlq' could not be found, or perhaps you do not have access to it. Please make sure the cluster ID is correct, and that you are a member of the organization that it belongs to."
               }
     patch:
       operationId: modifyCluster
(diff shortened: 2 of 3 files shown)


oponder:
To simplify things in the code, I considered just returning 404 for the `POST /v4/cluster` endpoint as well when the user doesn't belong to an organization. But I think it feels weird. How can the resource `/v4/cluster` be 'NOT FOUND'. That endpoint represents the collection of clusters the users has access to, and a POST to that endpoint means creating a new resource in that collection. It is not semantically reasonable for me to reply 'NOT FOUND' to attempts to act on such a collection.

In this case what is 'NOT FOUND' could be the organization the user passed in the `owner` field, but that is not the concern of this endpoint. This endpoint (the collection of clusters) definitely exists. Just the specific operation the user is trying to perform on it is invalid, and so we let him know with a 401.

xh3b4sd:
Based on your reasoning I think this is the right way to go. Implementation wise the suggestion looks straight forward.

marians:
@oponder When you write `/v4/cluster` or `/v4/cluster/`, do you actually mean `/v4/clusters/`?

oponder:
@marians yup, fixed

oponder:
Would appreciate a approval if we all think this is the right way to handle these responses. Also would implicitly assume that we apply the same pattern to other resources, though I'll make PR's similar to this one (up next is key pairs)

spec.yaml
```diff
@@ -195,10 +213,16 @@ paths:
                 "code": "RESOURCE_DELETION_STARTED",
                 "message": "The cluster with ID 'wqtlq' is being deleted"
```
xh3b4sd: Not your fault probably but here is a full stop missing at the end of the sentence.

spec.yaml
```diff
@@ -195,10 +213,16 @@ paths:
...
-          description: Error
+        "404":
+          description: Cluster not found
           schema:
             $ref: "definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "RESOURCE_NOT_FOUND",
+                "message": "The cluster could not be found."
```
xh3b4sd: FYI I like having information like a cluster ID in the response. 

```
The cluster with ID 'wqtlq' could not be found. 
```

marians:
I'll add my review in a bit

spec.yaml
```diff
@@ -89,10 +89,16 @@ paths:
                 "code": "RESOURCE_CREATED",
                 "message": "A new cluster has been created with ID 'wqtlq'"
               }
-        default:
-          description: Error
+        "401":
+          description: Permission Denied
```
marians: I would prefer `Permission denied` here (no title case), for consistency with `Cluster not found` etc.

marians: It makes sense to keep the `default` response everywhere, as this saves us from specifying every single possible response code, including `500` (Internal server error).

oponder: I dont really understand the point of 'default'.. why is it better? how does not using it force us to specify every single possible response code?

oponder: It ends up looking like this when I use 'default':
screen shot 2017-04-20 at 13 13 09

marians: Without "default" we would have to describe every possible response (status code) in every operation, to create a valid spec that could be used to create functioning clients.

If you use the spec to create a client, and the API operation returns a status code not contained in the spec (e.g. 500), and there is no "default" response for the operation, the client can't handle it.

With "default" we can just describe that there are responses with status codes other then the ones described explicitly here, all using the same response body format (V4GenericResponse). As `description` we can probably use something like `Other error` or maybe we can even omit `description`.

marians: To add to that, "default" is a way of communicating "there might be other response (error) types not described here".

oponder: Ah ok I get it now :D it's a client concern. Though in the documentation "default" doesn't mean that much to me and I didn't jump to that conclusion of "there might be other response types not described here"

So from the point of view of documenting and being kind of nice about it.. would it be ok if we specified the ones we know and care about (200, 401) and also make sure that they all have a default entry like before? I feel like that will 'read' nicer to the user looking at the docs.

Talked a bit about this in slack. But just documenting here.

marians: I totally agree that we should document the expected responses. I simply don't want to put work into duplicating an error 500 description for every operation.

oponder:
Ok cool! we've reached some consensus on how these responses should be. Implementation incoming today hopefully. Won't merge till implementation is done and deployed.

definitions.yaml
```diff
@@ -3,14 +3,13 @@ definitions:
   # Generic response model
   V4GenericResponse:
     type: object
-    required:
-     - code
-     - message
     properties:
       message:
         type: string
+        description: "A human readable message"
```
marians: FYI: The quotes are not necessary. Same further below.

oponder:
Only the `patch` responses do not do what they should according to this spec yet. Working on that now

oponder:
We did it \o/

oponder:
```
./kubectl cluster-info
Unable to connect to the server: x509: certificate signed by unknown authority

./kubectl cluster-info returned exit code 1

Action failed: ./kubectl cluster-info
```

Failed because of certificate not updated yet for this project on circleci.
Will take care tmw.

oponder:
There are no certificates to change to yet, because leaseweb g8s isn't up yet. 

Can I merge this even though the build is failing?

xh3b4sd:
You can simply delete the certificates of the old Leaseweb G8S and the build will work out. We have to add the new ones anyway later.

oponder:
ok, I have to update this to deploy with architect then

oponder:
on it

PR #15, opened 2017-04-20
Fixes in YAML structure; Adding validation tools

marians:
- `parameters.yaml` was missing a parent key at the beginning of the file, which has been added. References were updated. Also, unused items were deleted.
- To validate the concatenated spec YAML, a Makefile with `make validate` (for local use) and `make test` (for CircleCI use) has been introduced.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+spec-tmp.yaml
+spec-tmp.yamle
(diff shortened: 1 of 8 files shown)


oponder:
I had to `sudo pip install yamllint` before the `make validate` would work. I'm fine with that. 

Perhaps time for a requirements section in the README though.

marians:
@oponder README is updated now

PR #16, opened 2017-04-20
V4 Spec: getOrganizations

marians:
This PR contains the spec proposal for `GET /v4/organizations/`.


oponder:
@marians I'm going to implement this now, are you happy with the way it's specced? Since you never put it up for review, is it still in progress?

I don't have any comment. It looks good to me. Including the bit about admin's being able to list all organizations

marians:
@oponder The mechanics is still fine with me. Feel free to make this happen!

I'll change the description part a bit: This `The resulting list will contain all organizations the user is allowed to know about.` doesn't exactly make sense, as we every user is allows "to know" about an org's existence.

So for a normal user, the resulting list should contain all organizations the user is a member of.

For an admin, the list should simply contain all organizations.

marians:
FYI @oponder: Merged master, updated spec

oponder:
Thanks!

oponder:
part of: https://github.com/giantswarm/api/issues/338

oponder:
@marians: Technically your PR to merge. Listing orgs implementation is done

marians:
Nice! 🎱

PR #19, opened 2017-05-12
Conform to releng conventions

oponder:
In order to be consistent with other project's ways of doing things so far I have made some changes:

I have:
- renamed `api-spec-docserver` to just `api-spec`
- changed the ingress url from `apispec.g8s...` to `api-spec.g8s...`
- moved dockerfile to the root of the project
- moved kubernetes folder to the root of the project

In my opinion there is no need to call the image `api-spec-docserver` and it is nicer to have the url be consistent with the repo name.

This way the repo name, image, and url are all consistent with eachother.


oponder:
End result: 
http://api-spec.g8s.fra-1.giantswarm.io/fix-deploy/

oponder:
@marians Thanks!

PR #21, opened 2017-05-15
Update dockerfile, develop using docker

oponder:
Add's a Makefile target to start a preview server. This way we develop using the same image as what is deployed. 

This way we don't have a `docserver` and `redoc` folder with basically the same contents.

Updated nginx tag


oponder:
It passes quay's security scan now

PR #22, opened 2017-05-23
Spec creating and deleting orgs for /v4

oponder:


spec.yaml
```diff
@@ -299,3 +301,62 @@ paths:
...
+      summary: Create an organization
+      description: |
+        This operation allows a user to create an organization.
+      parameters:
+        - name: body
+          in: body
+          required: true
+          description: |
+            An ID for the organization to be created is the only required parameter.
+            This ID can only contain alpha-numeric characters, hyphens, underscores, and dots.
```
xh3b4sd: I am actually not sure about the format at this point. In such descriptions I think it makes more sense to provide some regular expression rule. I think on this level people understand it.

spec.yaml
```diff
@@ -299,3 +301,62 @@ paths:
...
+      responses:
+        "201":
+          description: Organization created
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "RESOURCE_CREATED",
+                "message": "A new organization has been created with ID 'OrgID'"
```
xh3b4sd: According to your API PR the message format is this. 

```
"The organization with ID '%s' has been created."
```

See https://github.com/giantswarm/api/pull/386/files#diff-004e9e81a905ac1c76d8c0e8fdb0b4a7R126.

You also miss the period at the end of the sentence.

spec.yaml
```diff
@@ -299,3 +301,62 @@ paths:
...
+          in: body
+          required: true
+          description: |
+            An ID for the organization to be created is the only required parameter.
+            This ID can only contain alpha-numeric characters, hyphens, underscores, and dots.
+          schema:
+            $ref: 'definitions.yaml#/definitions/V4AddOrganizationRequest'
+      responses:
+        "201":
+          description: Organization created
```
xh3b4sd: The description sometimes seems to be a sentence and sometimes not. Is this on purpose?

oponder:
I think it's not very common to see PUT used for creating. Even if it makes sense in REST semantics, and given the fact we supply the ID.

Furthermore I think the fact that we give organizations ID's like this is going to bite us soon.

It already means we can't really rename organizations (at least not without traversing all objects and finding places where that organization ID is referenced) and limits multitenancy (because of potential organization name collision) for our cloud offering. 

That said, I'll agree to disagree for now, spec it as a PUT and document the 409 case.

spec.yaml
oponder: Trying to follow the pattern already set up by other specced requests. If I got it wrong let me know.

spec.yaml
oponder: I'll fix that in the implementation. This is changing to a PUT also instead of a POST.

spec.yaml
oponder: Perhaps both a regex and the worded explanation?

Let me know if this is a blocker for you.

oponder:
rfr @marians . I changed it to a PUT, and am returning organization details now instead of a generic response.

spec.yaml
xh3b4sd: I don't know what is "right" here. I just see that it is not consistent.

spec.yaml
xh3b4sd: Not a blocker at all. Only personal opinion and thinking out loud.

parameters.yaml
```diff
@@ -6,3 +6,12 @@ parameters:
...
     description: Cluster ID
+
+  OrganizationIdPathParameter:
+    name: organization_id
+    in: path
+    required: true
+    type: string
+    description: |
+        An ID for the organization to be created is the only required parameter.
+        This ID can only contain alpha-numeric characters, hyphens, underscores, and dots.
```
marians: The actual requirements are more strict. It has to match `[a-z0-9_]{4,30}`, so no hyphens, dots, And it has to be unique.

See https://github.com/giantswarm/giantswarm/issues/15

oponder: That might be what we talked about in 2015, but I believe this is what the real validator is for org ids: https://github.com/giantswarm/companyd/blob/96668b4c7d398e7d59cfdaa7600bb12d3cc8ddbd/service/validator.go#L12

Unless there is some other validator up the chain somewhere.

oponder: screen shot 2017-05-24 at 13 42 31

You can see that apparently hyphens, dots, and even colons are ok

oponder: But since this is supposed to be about the spec. I am ok with writing what we _want_ it to be, and then fixing this in the correct microservice

oponder:
rfr @marians, I added an example, and edited the org id requirements based on the issue you linked. Though they are not in line with the actual implementation at the moment.

spec.yaml
```diff
@@ -299,3 +301,67 @@ paths:
...
+        - $ref: 'parameters.yaml#/parameters/OrganizationIdPathParameter'
+      responses:
+        "201":
+          description: Organization created
+          schema:
+            $ref: "definitions.yaml#/definitions/V4Organization"
+          examples:
+            application/json:
+              {
+                "id": "Acme",
```
marians: Please change `Acme`to `acme`

marians:
So far it isn't documented here that the creator automatically becomes a member of the created organization. Or did I miss it?

How is this handled when the admin user creates an org? I suppose the admin can't be an org member.

oponder:
Currently the admin behavior isn't specced or implemented for create and delete. That means right now, and admin won't be able to do anything special at these end points.

oponder:
part of: https://github.com/giantswarm/api/issues/338

PR #23, opened 2017-05-31
Specify that admin users are able to delete any orgs

oponder:
The admin user hasn't been given much love in v4 yet. For organization delete, the admin should be able to delete any org. This PR adds that 'rule' the spec.

There is no need to talk about the admin user in the create org part, because creating an org should behave the same way for normal users as well as admin users.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -348,7 +348,7 @@ paths:
       summary: Delete an organization
       description: |
         This operation allows a user to delete an organization that they are a member of.
-        Admin users, however, are able to delete organizations they are not a member of.
+        Admin users can delete any organization.
       parameters:
         - $ref: 'parameters.yaml#/parameters/OrganizationIdPathParameter'
       responses:
--- a/spec.yaml
+++ b/spec.yaml
@@ -347,7 +347,8 @@ paths:
         - organizations
       summary: Delete an organization
       description: |
-        This operation allows a user to delete an organization.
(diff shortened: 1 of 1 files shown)


oponder:
Turn out btw that this is already implemented because of the way the http authentication service works in `api`. 

There are some things wrong in my point of view with how it is implemented, but I'll try to word that and make an issue in API about it, and doesn't have anything to do with `api-spec`

For myself: What is wrong is that the admin token gets used as the user id further on in the chain. This means the admin token ends up in etcd as some kind of orphaned foreign key. For example when creating an organization as an admin, the user id to company mapping is stored like this in etcd:

```
/companyd/user_company/userid/orgid
/companyd/company_user/orgid/userid
```

This means that when an admin makes an org, we get:
```
/companyd/user_company/ADMINTOKEN/orgid
/companyd/company_user/orgid/ADMINTOKEN
```

ADMINTOKEN is not a valid userid, and our code doesn't mind invalid user ids when trying to list members that belong to an organization, so it doesn't cause an error. But I also think it isn't right to go around creating company-user relationships that are invalid, and that contain our admin token in them.

oponder:
I was wrong that the ADMINTOKEN gets stored in etcd.

What gets stored is the first part if the admin token before the 'colon'. (https://github.com/giantswarm/api/blob/master/kubernetes/deployment.yml#L56)

Still a bit 'wrong' that we end up creating invalid company-to-user relationships though

Issue #24, opened 2017-06-01
Support Scaling in API


oponder:
We need to discuss how scaling of worker nodes should work.

Related to: https://github.com/giantswarm/giantswarm/issues/1270

@puja @rossf7  @marians @oponder

marians:
- Scaling up and down should be done using simple patch requests.
  - We are aware of possible concurrency problems. This might be solved at a future point, e. g. using revision IDs.
- Auto scaling groups might take care of keeping the required amount of healthy nodes up. This isn't exposed via the API.
- By giving nodes unique labels, users can enforce removal of a specific node when required
- At first, removing a node will not drain the worker. Users will have to take care themselves. Our UIs should give appropriate warning.
- We need to restrict AWS clusters to using the same instance type for all nodes (as an intermediate step). We should have simple validation, throwing an issue when this restriction isn't met.

marians:
FYI: I removed the word "Update" from the title, as this here is about scaling only. "Updating" implies a change of the k8s version.

puja108:
btw, we should maybe also have a look at https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md to see if we could be compatible with the standard autoscaler somehow or at least not break certain standards

marians:
Implementation is tracked in https://github.com/giantswarm/api/issues/405. Closing this.

PR #27, opened 2017-06-02
Spec/docs for scaling clusters (adding/removing cluster nodes)

marians:
For https://github.com/giantswarm/api-spec/issues/24

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 spec-tmp.yaml
 spec-tmp.yamle
 kubernetes/secrets.yaml
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -71,7 +71,7 @@ definitions:
         properties:
           instance_type:
             type: string
-            description: EC2 instance type name
+            description: EC2 instance type name. Must be the same for all worker nodes of a cluster.
       memory:
         type: object
         properties:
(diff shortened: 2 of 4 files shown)


oponder:
Should we have some mention in here that when creating / scaling aws nodes, they must be all of the same instance type?

rossf7:
I think it would be good to mention it. We can remove once `aws-operator` supports multiple instance types for workers.

marians:
> Should we have some mention in here that when creating / scaling aws nodes, they must be all of the same instance type?

That's not what I understood. I understood that we create one Auto Scaling Group per instance type. If the user has three nodes already with three different instance types and adds a fourth node now with yet a different instance type, it should still work. Right?

puja108:
The first step will be a single ASG (so limit to 1 instance type), next will be ASG per instance type, the single ASG/instance is only a minor step in between

oponder:
I guess it depends on when we want to merge and publish this spec.

marians:
Excuse me, yes. I could have looked at my notes...

We should merge this ASAP, so of course I'll change it to inform about the restriction.

marians:
Content is now updated. If you like, please re-check.

marians:
Damn, I shouldn't have merged this already, as it isn't implemented yet...

marians:
Reverting this PR merge in https://github.com/giantswarm/api-spec/issues/33

PR #28, opened 2017-06-02
Idempotency documentation and header spec

marians:
For https://github.com/giantswarm/giantswarm/issues/1123

### DIFF
--- a/parameters.yaml
+++ b/parameters.yaml
@@ -15,3 +15,14 @@ parameters:
     description: |
         An ID for the organization to be created is the only required parameter.
         This ID must be unique and match this regular expression: ^[a-z0-9_]{4,30}$
+
+  XIdempotencyKeyHeader:
+    name: X-Idempotency-Key
+    in: header
+    type: string
+    required: false
+    description: |
+      Idempotency key. If given, the request is guaranteed to be executed only
+      once, even if the same request data including the same idempotency key is
+      submitted multiple times. The value must be between 32 and 64 characters
+      long. Find out more in the [idempotency](#idempotency) section.
--- a/spec.yaml
+++ b/spec.yaml
@@ -17,6 +17,26 @@ info:
 
     Feedback on the API as well as this documentation is welcome via `support@giantswarm.io` or on IRC channel [#giantswarm](irc://irc.freenode.org:6667/#giantswarm) on freenode.
 
+    ## Idempotency
+
+    Idempotency is the concept of giving a guarantee that a request is executed only once, if a few conditions are met.
+    This is currently only supported for [creating clusters](#operation/addCluster), but other operations will likely follow in the future.
+
+    The concept allows you to retry a request for creating a cluster in the case of an error, like a request timeout.
+
+    To enforce idempotency, your request must contain a header (`X-Idempotency-Key`) with a string value that you define yourself.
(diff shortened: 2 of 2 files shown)


parameters.yaml
```diff
@@ -15,3 +15,14 @@ parameters:
...
         This ID must be unique and match this regular expression: ^[a-z0-9_]{4,30}$
+
+  XIdempotencyKeyHeader:
+    name: X-Idempotency-Key
+    in: header
+    type: string
+    required: false
+    description: |
+      Idempotency key. If given, the request is guaranteed to be executed only
+      once, even if the same request data including the same idempotency key is
```
xh3b4sd: The request can be executed multiple times. This is the trick about this feature. It is more that we ensure to work towards a desired state, which at the end will be the same, no matter how often we try to reach the desired state. At some point we reach it and no further changes will be applied to reasources being created.

spec.yaml
```diff
@@ -17,6 +17,26 @@ info:
...
+    The concept allows you to retry a request for creating a cluster in the case of an error, like a request timeout.
+
+    To enforce idempotency, your request must contain a header (`X-Idempotency-Key`) with a string value that you define yourself.
+    The string must be between 32 and 64 characters long.
+    Only the characters `A-Z`, `a-z`, `0-9` and the hyphen `-` are allowed, however the string must neither start nor end with a hyphen.
+    For example, [UUID version 4](https://en.wikipedia.org/wiki/Universally_unique_identifier) is a useful way to create an idempotency key.
+
+    The API guarantees to execute only the first request under these conditions:
+
+    - All requests are issued by the same user
```
xh3b4sd: This is not implemented yet.

spec.yaml
```diff
@@ -17,6 +17,26 @@ info:
...
+    - The time between all requests is shorter than 24 hours
```
xh3b4sd: This is not implemented yet.

parameters.yaml
marians: I'll change the description to make that more obvious.

marians:
Description updated. Happy for a re-check.

marians:
@xh3b4sd I changed the long description of the idempotency concept. IMHO the short one describing the header is okay as it is. The context doesn't allow for much larger explanation.

spec.yaml
```diff
@@ -17,6 +17,25 @@ info:
 
     Feedback on the API as well as this documentation is welcome via `support@giantswarm.io` or on IRC channel [#giantswarm](irc://irc.freenode.org:6667/#giantswarm) on freenode.
 
+    ## Idempotency
+
+    Idempotency is the concept of making API calls repeatable, while guaranteeing that the state changes associated with the API call are only executed once.
+    Idempotency allows you to simply repeat certain API requests in case of an error.
+
+    Currently the only operation supporting idempotency is the [creation of clusters](#operation/addCluster), but other operations will likely follow in the future.
```
xh3b4sd: I checked the cluster-service and saw that we already do this for the deletion as well. See https://github.com/giantswarm/cluster-service/blob/ce61d2ef05081fc70f1810519f68fc9e0397447b/service/cluster/deleter/service.go#L116.

marians: That's good to know. I'll add that.

marians:
Changed. Merging.

Issue #26, opened 2017-06-02
Discuss: Listing users as admin


oponder:
For a while now I have wanted to be able to easily list all users, along with all their properties. So we can for example tell quickly whether or not they are expired.

I imagine it would live at:

`GET /v4/users/`

Do you think it is ok to have such an endpoint?

ping @marians

marians:
Yes, I think we should have the endpoint `GET /v4/users/`.

In general, I recommend to spec this so that the returned list items aren't complete, but sparse representations of the actual objects. This is how we do it with clusters as well. This means: we select which attribute of the member object are present in the list item. (And BTW adding an attribute later is never a breaking API change.)

xh3b4sd:
I am wondering why we need this now. Why don't we work on the user management story to improve how we actually do user management? There have been rumours we are going to play with dex and stuff. I see that we here now do changes towards our old userd, which I really do not want to do, because we will hopefully throw it away in the upcoming months.

oponder:
> Dex is NOT a user-management system, but acts as a portal to other identity providers through "connectors." 

We still need some place where we keep track of who is supposed to have access to our system, what orgs they belong to, etc. Dex just helps us identify them using something other than a password.

oponder:
And another point: I just don't want to have to ssh into machines and dig into etcd to do user management. This is just something that I think is a basic need for administrators. Even when we swap out userd with something else. 

I want this endpoint regardless of the underlying microservice or storage that we use.

So this technically is part of the user management story.

xh3b4sd:
This sounds like we have to stick with userd, which makes me super sad.

oponder:
We don't have to. I still have to see how far dex helps us. I think dex can replace userd, but not companyd.

oponder:
Done!

PR #30, opened 2017-06-03
/v4/users/ - listing users

oponder:
Related to: https://github.com/giantswarm/api-spec/issues/26

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -183,7 +183,7 @@ definitions:
         description: Email address of the user
 
   # One of the users in the array as returned by getUsers
-  V4GetUsersUser:
+  V4UserListItem:
     type: object
     properties:
       email:
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -181,3 +181,17 @@ definitions:
       email:
         type: string
         description: Email address of the user
+
+  # One of the users in the array as returned by getUsers
+  V4GetUsersUser:
+    type: object
+    properties:
+      email:
+        type: string
+        description: Email address of the user
+      created:
+        type: string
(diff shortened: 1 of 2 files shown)


definitions.yaml
```diff
@@ -181,3 +181,17 @@ definitions:
       email:
         type: string
         description: Email address of the user
+
+  # One of the users in the array as returned by getUsers
+  V4GetUsersUser:
```
marians: That name is a bit weird. Can we rename that to `V4UserListItem ` (according to `V4OrganizationListItem`)?

spec.yaml
```diff
@@ -41,6 +41,8 @@ tags:
       description: "User guide: Accessing Pods and Services from the Outside"
   - name: organizations
     description: Organizations are groups of users who own resources like clusters.
+  - name: users
+    description: A user represents a person that should have access to the Giant Swarm API. Users can belong to many groups, and are identified by e-mail address.
```
marians: `e-mail` -> `email`

spec.yaml
```diff
@@ -50,6 +52,42 @@ securityDefinitions:
...
 
 paths:
+  /v4/users/:
+    get:
+      operationId: getUsers
+      tags:
+        - users
+      summary: Get users
+      description: |
+        An admin only endpoint that returns a list of all users in the system.
```
marians: I could imagine opening this endpoint to non-admin users as well. It could then list all users the requesting user shares an organization with. But not now.

If we agree that this is likely, we should rephrase this along the lines of "Currently this endpoint/function is only available to users with admin permissions".

Issue #29, opened 2017-06-02
Question: How to approach add / remove members from organizations


oponder:
@marians how do you think we should approach adding and removing members from organizations?

Do you think we should use `PATCH /v4/organizations/:org-id/` with some json in the body representing the full final state of the members as the user wants it to be?

```
              {
                "members": [
                  {"email": "user1@example.com"},
                  {"email": "user2@example.com"},
                  {"email": "user3@example.com"},
                ]
              }
```

Meaning that user then has to supply the full list of members each time he wants to add one. (And the full list of members, minus the one he wants to remove, each time he wants to remove one.) But a benefit is that bulk operations are now possible.

Or do we go for something like:

`PUT /v4/organizations/:org-id/members/:email/`
`DELETE /v4/organizations/:org-id/members/:email/`

Meaning we leak personally identifiable information in our requests over the wire.

oponder:
Oh nevermind, I am apparently mistaken about the leaking information bit. URLs are encrypted too as part of https

marians:
I think we should use PATCH for all resource updating, so also in this case here. When `members` is the only attribute of the resource, then the PATCH payload will in fact represent the entire resource.

oponder:
> I think we should use PATCH for all resource updating

Agreed. But we could add the "/v4/organizations/:org-id/members/" resource, and then standard  `PUT` / `DELETE` becomes possible.

We're not forced to support PATCH for every field on every resource. We can make `members` an immutable field on the `organizations` resource if we want to. And introduce `members` as a resource.

That would be a lot simpler to implement right now, and means people wouldn't have to send a large list members each time they want to add or remove one. We'd also have to implement `revision id` to avoid concurrency issues. 

(i.e. someone else might add a member in between the time that the you fetched the list of all members, changed it, and then `PATCH`ed it back. Or basically the same situation we're going to face with the cluster scaling)

oponder:
Anyways I'm fine with implementing `PATCH` on the organizations resource. I just thought from a ux point of view its a bit nicer to have a members resource that we can `PUT` / `DELETE`

But then I realised if people are using happa or gsctl, then it doesn't matter too much. Though it is easier to automate, if some service down the line wants to add / remove a member from an organization, it effectively becomes a two request process (fetch the org, edit the list, patch the org)

Any final thoughts? I would still prefer having a `members` resource, but if you're against that, I can go with the `PATCH` on the organizations resource.

oponder:
[...]

PR #31, opened 2017-06-07
Spec for PATCH /v4/organizations/:org-id/

oponder:
Discussed here: https://github.com/giantswarm/api-spec/issues/29

I think there are a few more things to discuss though:

**1. What to do when the request is invalid?**
Currently I have specced it so that the entire patch operation fails. So if a member is being added that doesn't exist, or has an invalid email address, no  add/remove of members will happen at all, even if there might be other members in the patch request that are valid. Do we agree this is the right way to go about it? An alternative would be to just ignore invalid input.

**2. I think we need a new giant swarm response code:**
We have `UNKNOWN_ATTRIBUTE` and `IMMUTABLE_ATTRIBUTE` but that doesn't really match with what is going wrong here. What do you think of `INVALID_INPUT` with a http error code of `400` ?

ping @marians

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -452,6 +452,17 @@ paths:
         - organizations
       parameters:
         - $ref: 'parameters.yaml#/parameters/OrganizationIdPathParameter'
+        - name: body
+          in: body
+          required: true
+          schema:
+            properties:
+              members:
+                type: array
+                description: List of members that belong to this organization
+                items:
+                  $ref: 'definitions.yaml#/definitions/V4OrganizationMember'
+
       summary: Modify organization
       description: |
         This operation allows you to modify an existing organization. You must be
@@ -474,6 +485,16 @@ paths:
           description: Organization modified
           schema:
             $ref: 'definitions.yaml#/definitions/V4Organization'
+        "400":
+          description: Invalid input
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "INVALID_INPUT",
+                "message": "The organization could not be modified. (request is invalid for various reasons: 'asdf@' is not a valid email address, 'bob@example.com' does not exist)"
+              }
         "401":
           description: Permission denied
           schema:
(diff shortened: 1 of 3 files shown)


marians:
1. When part of the request is invalid (non-existing user), the entire request should fail and no changes should be applied. This way it's much easier for a client to understand what's been done and what hasn't.

2. `INVALID_INPUT` with 400 is fine with me. If you like, we could even be more specific for this case by using a code like `INVALID_USER`.

It would be nice to sort the responses in the spec.yaml by response code, ascending (default at the end).

oponder:
I think `INVALID_INPUT` should cover it for now, otherwise we'll be probably be adding `INVALID_CLUSTER`, `INVALID_KEYPAIR` and `INVALID_ORGANIZATION` too.

spec.yaml
```diff
@@ -446,6 +446,79 @@ paths:
...
+        - `members`: By modifying the array of members, members can be added or removed from the organization
+
+        The request body must conform with the [JSON Patch Merge (RFC 7386)](https://tools.ietf.org/html/rfc7386) standard.
+        Requests have to be sent with the `Content-Type: application/merge-patch+json` header.
+
+        The full request must be valid before it will be executed, currently this
+        means every member you attempt to add to the organization must actually
+        exist in the system. If any member you attempt to add is invalid, the entire
+        patch operation will fail, no members will be added or removed, and an error message
+        will explaining which members in your request are invalid.
```
marians: explaining -> explain

spec.yaml
```diff
@@ -446,6 +446,79 @@ paths:
...
+                  $ref: 'definitions.yaml#/definitions/V4OrganizationMember'
+
+      summary: Modify organization
+      description: |
+        This operation allows you to modify an existing organization. You must be
+        a member of the organization or an admin in order to use this endpoint.
+
+        The following attributes can be modified:
+
+        - `members`: By modifying the array of members, members can be added or removed from the organization
```
marians: or "added to or removed from" ?

oponder:
Thx for reviewing!

PR #32, opened 2017-06-08
Spec for GET /v4/clusters/

marians:
This adds the spec for `GET /v4/clusters/`.

Minor changed in the definition of `V4GenericResponse` to improve the readability in the rendered HTML docs.

**To be merged after implementation**

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -115,7 +115,7 @@ paths:
         - clusters
       summary: Get clusters
       description: |
-        This operation allows to fetch a list of clusters.
+        This operation fetches a list of clusters.
 
         The result depends on the permissions of the user.
         A normal user will get all the clusters the user has access
--- a/spec.yaml
+++ b/spec.yaml
@@ -109,6 +109,51 @@ paths:
             $ref: "definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/clusters/:
+    get:
+      operationId: getClusters
+      tags:
+        - clusters
+      summary: Get clusters
+      description: |
+        This operation allows to fetch a list of clusters.
+
+        The result depends on the permisisons of the user.
+        A normal user will get in return all the clusters the user has access
+        to, via organization membership.
+        A user with admin permission will receive a list of all existing
+        clusters.
+
+        The result array items are sparse representations of the cluster objects.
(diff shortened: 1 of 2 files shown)


spec.yaml
```diff
@@ -109,6 +109,51 @@ paths:
             $ref: "definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/clusters/:
+    get:
+      operationId: getClusters
+      tags:
+        - clusters
+      summary: Get clusters
+      description: |
+        This operation allows to fetch a list of clusters.
```
oponder: You often take this sentence structure, but I think its grammatically incorrect.

It allows (SOMETHING) to do something.

`It allows to do something` isn't right I think.

marians: Hmm. I pretty much copied myself from other descriptions. Let's ask @rossf7 or @JosephSalisbury for native speaker help. Is there a problem with that phrase? If yes, what's a better way?

JosephSalisbury: My gut feeling is that I would say "this operation fetches", but I don't see anything wrong with this.

spec.yaml
```diff
@@ -109,6 +109,51 @@ paths:
...
+
+        The result depends on the permisisons of the user.
```
rossf7: Typo on permissions

spec.yaml
```diff
@@ -109,6 +109,51 @@ paths:
...
+        A normal user will get in return all the clusters the user has access
```
rossf7: Super nitpicky but I think just "will get all the clusters" is easier to read.

spec.yaml
rossf7: Agree on fetches and 2 other minor issues but otherwise looks good :)

marians: @rossf7 As you reviewed already, but didn't comment here: Do you see a problem with the phrase `This operation allows to fetch a list of clusters.` ?

marians:
Thanks for you eyeballs! Changes made based on reviews.

spec.yaml
rossf7: @marians when I said I agree on fetches. I meant I agree with @JosephSalisbury ;) "this operation fetches" reads better to me.

PR #34, opened 2017-06-09
Spec for scaling clusters

marians:
**To be merged when implemented**

This PR applies the same changes as https://github.com/giantswarm/api-spec/pull/27, but this time, it should remain unmerged until ready.

TODO

- [x] Inform that scaling is not available on KVM based clusters.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -328,7 +328,6 @@ paths:
         when submitting a PATCH request, you ensure that the right instance type
         is used automatically.
 
-        Note for clusters on AWS: When adding worker nodes to a cluster, they must use the same `instance_type` as the existing workers.
       responses:
         "200":
           description: Cluster modified
--- a/spec.yaml
+++ b/spec.yaml
@@ -241,19 +241,92 @@ paths:
       tags:
         - clusters
       parameters:
+        - name: body
+          in: body
+          required: true
+          description: Merge-patch body
+          schema:
+            $ref: 'definitions.yaml#/definitions/V4ModifyClusterRequest'
         - $ref: 'parameters.yaml#/parameters/ClusterIdPathParameter'
       summary: Modify cluster
       description: |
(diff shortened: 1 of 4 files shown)


spec.yaml
```diff
@@ -244,9 +246,16 @@ paths:
       description: |
         This operation allows to modify an existing cluster.
 
-        __Note:__ This operation currently requires special permission. Also, currently only the `owner` attribute can be changed.
+        The following attributes can be modified:
+
+        - `owner`: Given that the requesting user has admin permissions, this can be used to change a cluster's ownership.
+        - `workers`: By modifying the array of workers, nodes can be added or removed to scale the cluster's capacity.
+        When adding nodes, the node definition has to conform the specification as detailed in the [cluster creation](#operation/addCluster) operation.
```
rossf7: Nitpick: "to the spec"

marians:
Related to #24

marians:
FYI: The description of `modifyCluster` is now changed so that only removing nodes is possible, as this will be the first scaling type to be implemented.

oponder:
> FYI: The description of modifyCluster is now changed so that only removing nodes is possible, as this will be the first scaling type to be implemented.

This should be "only adding nodes is possible" right? That's what we agreed on implementing first.

marians:
Correct, thanks!

marians:
FYI: I just added

- a missing request body definition for the patch call
- more documentation on the little bits and details to note when scaling.

Happy for you @oponder @rossf7 noticing.

marians:
Ah, and yes, the spec is no longer limited to only adding nodes.

marians:
We decided in the product meeting to merge this ASAP in order to make the spec/docs generally available. We'll have to inform KVM customers that they cannot use this yet.

I'll prepare another change that informs about the KVM restriction.

PR #36, opened 2017-06-19
Spec DELETE /v4/users/{email}/

oponder:
This specs deleting users.

EDIT: something went wrong with my hub client, and it pushed this with an totally unrelated PR message)

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -151,7 +151,6 @@ paths:
           description: Error
           schema:
             $ref: "definitions.yaml#/definitions/V4GenericResponse"
-  }
 
   /v4/clusters/:
     get:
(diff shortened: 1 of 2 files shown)


spec.yaml
```diff
@@ -108,6 +108,50 @@ paths:
...
+      responses:
+        "200":
+          description: User deleted
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "RESOURCE_DELETED",
+                "message": "The user  with email 'bob@example.com' has been deleted."
```
puja108: `user  with` -> `user with`

xh3b4sd: There is also a double space.

PR #43, opened 2017-08-17
Spec creating users

oponder:
I want a way to create users in the new /v4/ paradigm. This will let me add validations and sanitization that are missing currently at the v1 endpoint.

It also lets us get closer to a point where we can remove all legacy from api, which is something I really want to do.

I know we intend to replace user management entirely in the short/medium term, but I don't want to have any more tech debt and confusion caused by things we know but are not fixing.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -388,3 +388,16 @@ definitions:
             version:
               type: string
               description: Version number of the component
+
+  V4CreateUserRequest:
+    type: object
+    required:
+      - password
+    description: Request model for creating a new user
+    properties:
+      password:
+        type: string
+        description: A Base64 encoded password
+      expiry:
+        type: string
+        description: The date and time when this account will expire
--- a/spec.yaml
+++ b/spec.yaml
@@ -218,7 +218,7 @@ paths:
             application/json:
               {
                 "code": "RESOURCE_ALREADY_EXISTS",
-                "message": "The user could not be created. (invalid input: email 'bob@example.com' already exists.)"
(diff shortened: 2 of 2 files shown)


spec.yaml
```diff
@@ -109,6 +109,58 @@ paths:
...
+              }
+        "400":
+          description: User already exists
+          schema:
+            $ref: "definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "RESOURCE_ALREADY_EXISTS",
+                "message": "The user could not be created. (invalid input: email 'bob@example.com' already exists.)"
```
xh3b4sd: The full stop at the end of the error message in the braces is not right IMO.

definitions.yaml
```diff
@@ -226,3 +226,16 @@ definitions:
...
+
+  V4CreateUserRequest:
+    type: object
+    required:
+      - password
+    description: Request model for creating a new user
+    properties:
+      password:
+        type: string
+        description: A Base64 encoded password
```
marians: Not sure about the base64 encoding, as it's not a solid standard. See also my comments on https://github.com/giantswarm/api-spec/pull/49

oponder:
Got any strong feelings about the base64 encoding? I am happy to adjust that, but I finally have an implementation for this now.

definitions.yaml
puja108: not sure, but it seems to work quite well in k8s

marians:
No more strong feelings. The fact that it works well for kubernetes secrets is convincing.

PR #45, opened 2017-08-21
Add cn_prefix and o to keypair requests (RBAC)

oponder:
@puja, @marians, and I discussed this last Thursday, that it was becoming important in the short term to be able to set the Common Name of certificates we generate during key pair creation. Since it allows for setting more granular RBAC permissions per certificate instead of granting full admin rights each time.

This PR proposes a way to achieve that for the end user.

Implementation considerations:

We want to allow only a prefix of the Common Name of the certificate to be defined by the user. This lets us avoid collisions with other k8s components, as well as ensures users won't be able to make certificates that have no chance of working, since the K8s API will reject certificates that aren't a subdomain of the clusterdomain.

To avoid collisions with other k8s components, user certificate's common names should be a subdomain of `user.clusterdomain` like: `oliverponder_giantswarm_io.user.clusterdomain`

While we're at it, it would also be nice to be able to set the O of the certificate subject. 

Relevant info:
https://kubernetes.io/docs/admin/authentication/#x509-client-certs

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -130,7 +130,7 @@ definitions:
         cn_prefix:
           type: string
           description: The common name prefix of the certificate subject.
-        o:
+        certificate_organizations:
           type: string
           description: The certificate subject's `organization` fields.
 
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -127,6 +127,12 @@ definitions:
         create_date:
           type: string
           description: Date/time of creation
+        cn_prefix:
+          type: string
+          description: The common name prefix of the certificate subject.
+        o:
+          type: string
+          description: The certificate subject's `organization` fields.
 
   # Add key pair request
   V4AddKeyPairRequest:
@@ -141,6 +147,12 @@ definitions:
         type: integer
         format: int32
         description: Expiration time (from creation) in hours
+      cn_prefix:
+        type: string
+        description: The common name prefix of the certificate subject.
+      o:
+        type: string
+        description: This will set the certificate subject's `organization` fields. Use a comma seperated list of values.
 
   V4AddKeyPairResponse:
     type: object
--- a/docserver/index.html
+++ b/docserver/index.html
@@ -16,6 +16,6 @@
   </head>
   <body>
     <redoc spec-url='yaml/spec.yaml'></redoc>
-    <script src="https://rebilly.github.io/ReDoc/releases/v1.17.0/redoc.min.js"></script>
+    <script src="https://rebilly.github.io/ReDoc/releases/latest/redoc.min.js"></script>
   </body>
 </html>
--- a/docserver/index.html
+++ b/docserver/index.html
@@ -16,6 +16,6 @@
   </head>
   <body>
     <redoc spec-url='yaml/spec.yaml'></redoc>
-    <script src="https://rebilly.github.io/ReDoc/releases/latest/redoc.min.js"></script>
+    <script src="https://rebilly.github.io/ReDoc/releases/v1.17.0/redoc.min.js"></script>
   </body>
 </html>
--- a/spec.yaml
+++ b/spec.yaml
@@ -523,7 +523,7 @@ paths:
           to set up role-based access controls.
 
 
-        - `o`: This will set the certificate's `organization` fields. Use a comma seperated list of values.
+        - `certificate_organizations`: This will set the certificate's `organization` fields. Use a comma seperated list of values.
           The Kubernetes API will use these values as group memberships.
 
         __Note:__ The actual credentials coming with the key pair (key, certificate) can only be accessed once, as the result of the `POST` request that triggers their creation. This restriction exists to minimize the risk of credentials being leaked. If you fail to capture the credentials upon creation, you'll have to repeat the creation request.
--- a/spec.yaml
+++ b/spec.yaml
@@ -507,6 +507,25 @@ paths:
 
         In addition to the credentials itself, a key pair has some metadata like a unique ID, a creation timestamp and a free text `description` that you can use at will, for example to note for whom a key pair has been issued.
 
+        ### Customizing the certificate's subject for K8s RBAC
+
+        It is possible to set the Common Name and Organization fields of the generated certificate's subject.
+
+        - `cn_prefix`: The certificate's common name uses this format: `<cn_prefix>.user.<clusterdomain>`.
+
+          `clusterdomain` is specific to your cluster and is not editable.
+
+          The `cn_prefix` however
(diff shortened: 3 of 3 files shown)


oponder:
@xh3b4sd did you work mostly on the keypair creation and vault integration? Can you shed some light on if this spec seems achievable from that point of view?

puja108:
I aded @rossf7 as he has recently migrated Vault to the newest version which supports this functionality (the Os)

oponder:
> If necessary the cn_prefix can be used in kubernetesd when creating the cert TPOs.

I'm a bit confused about the changes needed to `certificatetpr` and `kubernetesd`

I don't think that'll be needed. If my mental model is right, kubernetesd is making the cluster, and the certificates for components of the cluster right?

The `cn_prefix` shouldn't appear in any of those certificates. The `cn_prefix` part is something the user can choose to set per certificate that he creates and is intended to identify a single user.

As far as I _knew_ the user made keypairs aren't TPO's. I might be wrong though, so please correct me if I misunderstand the situation.

puja108:
@oponder @rossf7 there's two things we need, and a lot of the requirements and changes are similar:
1. User should be able to get `<username>.user.clusterdomain` as CN and any O they need.
2. k8sd and and cluster-service should be able to generate certs with any O (we especially need `system:masters` and things like `system:node`, `system::kube-controller-manager` etc) as well as CNs that describe the component the cert is used by. For latter we can either use `<componentname>.clusterdomain` (like we already have with `api.`) or we use `<componentname>.component.clusterdomain` or something else for the `component` part in that domain.

So somewhere in the chain it should be possible to choose any CN and any O. The limitation to `user` subdomain can be pretty far up the stack (e.g. API + UIs only).

rossf7:
@oponder you're right the user key pairs aren't TPOs. I got sidetracked because there is a lot of overlap as @puja108 says.

We will need the changes to `certctl` to set the `o` for both of these requirements. As it's our library for integrating with Vault.

oponder:
@puja: understood. But I was talking about and responding to ross's use of `cn_prefix` specifically.

For what you're talking about, "the ability to choose any O and CN" would probably land in `certctl` since that is what makes the api call to vault.

oponder:
Cool, i think we're on the same page :D

oponder:
@puja: for the stuff about k8sd and changes to the certs that components are getting should be moved to a separate issue I think. Though its good to keep in mind this requirement. We shouldn't try to satisfy that here.

As long as `certctl` can set `O`, we should be good to iterate further on changes needed in `k8sd` and the operators. Could make an issue to describe/track those changes further?

oponder:
Here's my view of the required changes and what components will be affected.

1. `api`
    - add `CNPrefix` and `Organization` to the request struct in the keypair creator endpoint and 
    service. Pipe it through to the call to `cluster-service` using `clusterclient`
    - Add logic for setting the default value of `CNPrefix` in case the user left it blank.
2. `clusterclient`
    - add `CNPrefix` and `Organization` to the create keypair request struct to be able to achieve point 1 
3. `cluster-service`
    - use what was received for `CNPrefix` and prepend it to what was there previously for `CommonName`. Add `Organization`
4. `certctl`
    - add `Organization` as a setting in the call it makes to the `Vault API`. `CommonName` is already there.

I think I can handle all the changes, if it's just a matter of making the right call to `vault` from `certctl`.

puja108:
@oponder LGTM

Agree, that the k8sd functionality is a different ticket and doesn't go over API.

point 4 might be tricky and you might need to sync with @rossf7 there, IIRC Os in Vault belong to the "role", and currently we're using a single role in Vault, so current structure is:
`PKI - role - all certs`. It might be that we need to create a role per O (or actually combination of Os) in which all certs that need that (combination of) O get created. So it would be:
`PKI - role1 (O: foo, bar) - certs with O foo, bar`
`PKI - role2 (O: foo) - certs with O foo`
...

rossf7:
@puja108 Thanks for the pointer on the roles. I agree point 4 is tricky and I think needs more `certctl` changes.

@oponder Here is where the certctl PKI service creates the single role. So we'd need to be able to create multiple roles and maybe add roles to an existing PKI? 

https://github.com/giantswarm/certctl/blob/master/service/pki/service.go#L233

Hope that helps let me know if you want to sync more.

oponder:
Yeah I might need to do the `certctl` part in pair programming, or otherwise get some help with it.

But from what I'm reading, and from what I see in the code @rossf7 linked, it is possible to create role on the fly.

So I'm thinking I'd add similar code to what is in the `pki service` to the `cert-signer service`. Where it'll try to create a role if it doesn't exist yet, and then use that role to issue the certificate.
(https://github.com/giantswarm/certctl/blob/master/service/cert-signer/cert_signer.go#L54)

xh3b4sd:
Looks like you guys have it sorted. My two cents. I noticed we run Vault `0.6.4` on `lycan`. We have to make sure Vault is up to date with these requirements in all installations. Maybe SIG updates or SIG host cluster can organize this.

definitions.yaml
```diff
@@ -141,6 +147,12 @@ definitions:
         type: integer
         format: int32
         description: Expiration time (from creation) in hours
+      cn_prefix:
+        type: string
+        description: The common name prefix of the certificate subject.
+      o:
```
xh3b4sd: Would it make sense to make our public API more elaborative? Don't know if this would be better UX than a single letter.

oponder: Yeah, I was thinking about this. `o` feels super weird.

I thought to do:

`certificate_common_name_prefix` and `certificate_organizations`

or

`common_name_prefix` and `organizations`

But not sure. What do you think @puja108

xh3b4sd: `certificate_organizations` and `common_name_prefix` works for me. It is self explaining and aligns with the style we use so far.

puja108: both ok for me, you decide :)

puja108:
I was thinking today about a critical path, and as customers would like to have this rather sooner than later, we could start from the bottom and enable it in certctl first, so that we can at least already manually create the keypairs for them until they get the functionality to do it themselves. @oponder WDYT?

oponder:
Yup, im starting with certctl. Was hacking on it today a bit already, and will continue hacking tonight. I think I can have something up for review by tmw morning.

oponder:
I've change it from the key for setting the organizations subject from `o` to `certificate_organizations` and key `cn_prefix` as is.

@marians fyi: if you're around and want to give a final look at this spec, I'd appreciate it. I think you have good reasoning about the names of keys for things in the API.

In short summary: we're adding `cn_prefix` and `certificate_organizations` fields to the key pair create request.

PR #49, opened 2017-09-21
Add specs for creating / deleting auth tokens

oponder:
Having this in /v4/ will help us get closer to removing all legacy from API.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -504,9 +504,6 @@ definitions:
       owner:
         type: string
         description: Name of the organization owning the cluster
-      release_version:
-        type: string
-        description: The semantic version number of this cluster
 
   # A cluster array item, as return by getClusters
   V4ReleaseListItem:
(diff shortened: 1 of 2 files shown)


oponder:
These auth token routes can then become deprecated. 

The OIDC flow is quite different, it requires a redirect to start the flow, so I Imagine it would live at something like `/v4/oidc/login`

I even think `/v4/oidc/login` and `/v4/auth-tokens` can live together for a while.

definitions.yaml
```diff
@@ -241,3 +241,26 @@ definitions:
...
+        description: Your password as a base64 encoded string
+
+  # A response to a successful auth token request
+  V4CreateAuthTokenResponse:
+    type: object
+    properties:
+      auth-token:
+        type: string
+        description: The newly created API token
+      code:
```
marians: The successful response does not need a `code`.

Error responses use the `V4GenericResponse` type which has a code.

oponder:
@marians: Regarding the `base64` discussion, I wouldn't want to make changes here. I don't think there is an immediate security concern, and I envision this endpoint not having a very long lifetime anyways as we start to further flesh out our other authentication solutions.

marians:
@oponder Fine with me. base64 seems good enough.

oponder:
@marians can I get your approve, if you are ok with base64? also noticed (thanks t @tuommaki that I had put a hyphen in the json response field name for the auth token, fixed that now as well.

oponder:
Spec has been implemented.

definitions.yaml
```diff
@@ -504,9 +504,6 @@ definitions:
       owner:
         type: string
         description: Name of the organization owning the cluster
-      release_version:
-        type: string
-        description: The semantic version number of this cluster
```
marians: @oponder I just noticed that this PR has removed The `release_version` attribute from the V4ClusterListItem model. That doesn't seem right. Could you please provide a follow-up to fix that?

oponder: Oops, thats some mistake I must've done while merging. Fixing.

PR #52, opened 2017-10-24
Exposing guest cluster metrics in the API

marians:
Towards https://github.com/giantswarm/giantswarm/issues/1959

### Design notes

- The resource users deal with is the metric, hence the route `/v4/clusters/{cluster_id}/metrics/`, and hence the top level response items are metrics.
- We attempt to deliver all metrics in one response.
- We do not expose "static" metrics like the number of CPU cores available or the limit of pods per worker. These are supposed to be exposed in other responses, like installation info or cluster details (both TODO).
- We attempt to be consistent when it comes to the aggregation logic from node level values to the cluster level. The cluster level metric is always the sum of all according worker node values.

### Open questions

- What if we have a second metric for network transfer, using packets instead of bytes/second as a unit? Should we change the name from e. g. `network_tx` to `network_tx_bytes`?


spec.yaml
```diff
@@ -523,6 +525,74 @@ paths:
...
+        - `cpu_used`: Number of CPU cores used, as a floating point number. This
+        metric has no unit.
+        - `ram_used`: The amount of memory used in the node/cluster.
+        - `pod_count`: Number of pods running on the node/cluster. This metric
+        has no unit.
+        - `container_count`: Number of containers running in the node or
+        cluster. This metric has no unit.
+        - `node_storage_used`: Amount of worker disk storage used on the
+        node/cluster.
+        - `network_rx`: Incoming network transfer rate.
```
JosephSalisbury: What does `rx` mean? And `tx`?

definitions.yaml
```diff
@@ -182,6 +182,74 @@ definitions:
...
+    type: object
+    properties:
+      container_count:
+        description: Number of containers running
+        $ref: '#/definitions/V4MetricsBundle'
+      pod_count:
+        description: Number of pods running
+        $ref: '#/definitions/V4MetricsBundle'
+      cpu_used:
+        description: CPU usage as a value between 0 and 1
```
JosephSalisbury: Later we have:
> Number of CPU cores used, as a floating point number

I don't follow how the number of cpu cores is between 0 and 1

marians: That is fixed now.

spec.yaml
marians: Rx and Tx are common symbols for received and transmitted traffic.

oponder:
Am I assuming rightly that this is set up like desmotes was, to keep things easy to implement since we already have some code basically?

The thing about the way this is output is that it requires quite some work from the client to create a picture of all the metrics for a given cluster or worker.

It feels like the data structure is a bit 'inverted'. It would make more sense for me if things were like:

```
#psuedoJSON

cluster
   - container_count
   - pod_count
   - ...
workers: [
  {
   - container_count
   - pod_count
   - ...
   },
   {
   - container_count
   - pod_count
   - ...
   },
]
```

But since we already have the code that deals with the suggested structure, I am fine with this.

That said, since this is something we also want to share with customers, I think the sample structure above is easier and friendlier to reason with.

marians:
After talks with @oponder, changed this:

- general structure/grouping changed
- Removed "cluster" key, as it would contain simple aggregates which could be done by clients as well
- Removed unit info, as this is not expected to change and we can give that info in docs

marians:
FYI: I updated the PR to work around a documentation issue. ReDoc (which we use at https://docs.giantswarm.io/api/ ) does not parse the response model for the getClusterMetrics operation correctly. As a result, the descriptions of the individual metrics were not shown. I moved those into the description of the operation as markdown content.

marians:
It's implemented 🚀

PR #53, opened 2017-10-26
Info endpoint

marians:
Spec for https://github.com/giantswarm/api/issues/282

The goal here is to provide a first iteration of an info endpoint. The list of details doesn't have to contain everything we can think of, but is mostly informed by the use case we know about with certainty.

### Design notes

- The sub keys `min`, `max`, `default`, and `options` are part of a recurring scheme, however not all info entries can provide them.

### Open

- Authenticated vs. unauthenticated access: Do we want/have to expose some info details even to anonymous users?

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -73,22 +73,23 @@ paths:
 
         ```json
         {
-          "installation_name": "shire",
-          "provider": "aws",
-          "datacenter": "eu-central-1",
-          "workers_per_cluster": {
-            "max": 20,
-            "default": 3
+          "general": {
+            "installation_name": "shire",
+            "provider": "aws",
+            "datacenter": "eu-central-1"
           },
-          "worker_instance_type": {
-            "options": [
-              "m3.medium", "m3.large", "m3.xlarge"
-            ],
-            "default": "m3.large"
-          },
-          "cpu_cores_per_worker": null,
-          "ram_size_gb_per_worker": null,
-          "storage_size_gb_per_worker": null
+          "workers": {
+            "count_per_cluster": {
+              "max": 20,
+              "default": 3
+            },
+            "instance_type": {
+              "options": [
+                "m3.medium", "m3.large", "m3.xlarge"
+              ],
+              "default": "m3.large"
+            }
+          }
         }
         ```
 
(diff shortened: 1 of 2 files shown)


oponder:
Seems like we're speccing to include all keys, even if they are irrelevant for a certain installation. 

So the spec states that the info endpoint would return `null` for `worker_instance_type` on `kvm` installations. 

Wouldn't it be nicer to omit keys that are irrelevant for a given installation?

marians:
@oponder I think you're right. Removing null infos would make more sense.

Do you have an opinion regarding anonymous access? Should we expose any infos to unauthenticated users?

oponder:
My gut instinct says do not expose anything. I am even uncomfortable with the `go_version` we expose at the "`root` api endpoint of our api.

```
{
description: "This is the API Gateway to the rest of the Giant Swarm services.",
git_commit: "403dd0bfe39cfc2cc8958d84609ccf0b202f453b",
go_version: "go1.9.1",
name: "api",
os_arch: "linux/amd64",
source: "https://github.com/giantswarm/api"
}
```

oponder:
There could be a distinction made though between authenticated users and admin users.

We could show admins some more info. Perhaps total number of clusters, users, and orgs? Though that can all be found out as well by doing calls to `/v4/clusters/`, `/v4/users/`, `/v4/organizations/`.

So right now a YAGNI feature; but something to keep in mind as a possibility.

spec.yaml
```diff
@@ -54,6 +56,80 @@ securityDefinitions:
...
+          },
+          "worker_instance_type": {
+            "options": [
+              "m3.medium", "m3.large", "m3.xlarge"
+            ],
+            "default": "m3.large"
+          },
+          "cpu_cores_per_worker": null,
+          "ram_size_gb_per_worker": null,
+          "storage_size_gb_per_worker": null
```
xh3b4sd: I personally do not like this. It is nice to have a flat object but the naming of the object's properties is really not beneficial. I personally prefer having structured objects where properties are grouped based on their relationships or scopes. For instance it would make a lot of sense to me to have a separate block for all worker information.

marians:
@oponder With my question regarding exposing some info for anonymous users I also had happa in mind. There is this hard-coded part in happa's HTML:

```javascript
var config = {
  apiEndpoint: 'https://api.g8s.gauss.eu-central-1.aws.gigantic.io',
  passageEndpoint: 'https://passage.g8s.gauss.eu-central-1.aws.gigantic.io',
  desmotesEndpoint: 'https://desmotes.g8s.gauss.eu-central-1.aws.gigantic.io',
  intercomAppId: 'ictssdcu',
  domainValidatorEndpoint: 'http://localhost:5002',
  createClusterWorkerType: 'aws',
  environment: 'kubernetes',
  version: '0.1.22+git'
}
```

One attribute here overlaps with the info endpoint draft: `domainValidatorEndpoint` is the `provider` attribute.

marians:
With the last commit I grouped the info under a "general" and a "workers" key. This allowed for some simpler attribute naming.

marians:
Also removed `null` value keys, as @oponder suggested.

oponder:
> createClusterWorkerType  is the provider attribute.

this is still only something that authenticated users need to know. It determines the type of screen that the user sees when on the create cluster screen. So not something I need to know in the 'login' screen, before a user is authed.

marians:
@oponder "this is still only something that authenticated users need to know" > True

marians:
Happy for a re-review @oponder @xh3b4sd

oponder:
Implementation is in progress. I pondered long about the path that @xh3b4sd suggested, I started making the 'info service' and it really felt that it was too over engineered, so the implementation I am making is coming in `cluster-service` itself.

Making another service that just queries the k8s api, which will then query etcd to get the number of nodes in the cluster, and store that in the tpr storage for caching (which is also etcd) didn't make sense to me. 

For the expected use of this new endpoint, this much optimisation and a separate service isn't going to be needed in my opinion. How often do people create clusters anyways? Why take something from etcd, to put it back into etcd again?

xh3b4sd:
FYI we will take it out of cluster-service again anyway. Sooner or later. I am fine adding it to cluster-service now to not have the overhead with wiring a new microservice. I took this shortcut myself with the release related business logic. Anyway, fetching the number of nodes will not remain the only requirement and this service will grow and will have its own place. Lets also ignore etcd from the picture here. We need new functionality which has its own scope and own quirks and having a separate microservice for this makes total sense.

PR #54, opened 2017-11-06
Releases endpoint API spec

marians:
Part of https://github.com/giantswarm/giantswarm/issues/2022

This spec change introduces an endpoint that allows the API to inform users/clients which versions of guest cluster software/config are available both for creation of new guest clusters and for upgrades of existing guest clusters.

Status: Will be merged when implemented

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -792,27 +792,31 @@ paths:
                   "components": [
                     {
                       "name": "kubernetes",
-                      "version": "1.5.8"
+                      "version": "1.5.8",
+                      "description": "Security fixes"
                     },
                     {
                       "name": "calico",
-                      "version": "0.9.1"
+                      "version": "0.9.1",
+                      "description": "Security fixes"
                     }
                   ],
                   "deprecated": true
                 },
                 {
-                  "version": "2.8.3",
+                  "version": "2.8.4",
                   "timestamp": "2017-11-11T12:24:56.59969Z",
                   "description": "Minor update of calico, fixing a bug.",
                   "components": [
                     {
                       "name": "kubernetes",
-                      "version": "1.7.3"
+                      "version": "1.7.3",
+                      "description": null
                     },
                     {
                       "name": "calico",
-                      "version": "1.1.0"
+                      "version": "1.1.1",
+                      "description": "Bugfix"
                     }
                   ],
                   "deprecated": false
--- a/spec.yaml
+++ b/spec.yaml
@@ -100,7 +100,8 @@ paths:
         - users
       summary: Delete user
       description: |
-        Deletes a users in the system. Currently this endpoint is only available to users with admin permissions.
+        Deletes a users in the system. Currently this endpoint is only available
+        to users with admin permissions.
       responses:
         "200":
           description: User deleted
--- a/spec.yaml
+++ b/spec.yaml
@@ -46,7 +46,17 @@ tags:
   - name: users
     description: A user represents a person that should have access to the Giant Swarm API. Users can belong to many groups, and are identified by email address.
   - name: releases
-    description: A release is a software bundle that constitutes a cluster
+    description: |
+      A release is a software bundle that constitutes a cluster.
+
+      Releases are identified by their
+      [semantic version number](http://semver.org/) in the `MAJOR.MINOR.PATCH`
+      format.
+
+      A release provides _components_, like for example kubernetes. For each
+      release the contained components are listed. Changes in components are
+      detailed in the _changelog_ of a release.
+
 securityDefinitions:
   token_auth:
     description: Currently the only supported authentication scheme is passing an auth token via the `Authorization` header with a value of the format `giantswarm <token>`.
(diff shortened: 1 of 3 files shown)


marians:
@xh3b4sd Please take a look.

I am using the term "release" here after some thinking, as it's a quite common term and should tell what this is about. Open for debate.

spec.yaml
```diff
@@ -759,3 +762,59 @@ paths:
...
+            type: array
+            items:
+              $ref: "definitions.yaml#/definitions/V4ReleaseListItem"
+          examples:
+            application/json:
+              [
+                {
+                  "version": "1.14.9",
+                  "timestamp": "2017-09-21T08:14:03.37759Z",
+                  "description": "Minor updates of kubernetes and calico, both for security fixes.",
```
xh3b4sd: We have a list of changelogs available. I think it would be good to have then a list of descriptions. Our internal code requires to have one changelog (description) for each component that changed. See

- https://github.com/giantswarm/versionbundle/blob/efdbdae719c56ab4880f8fe56643d60a76fc19e5/changelog.go- https://github.com/giantswarm/versionbundle/blob/efdbdae719c56ab4880f8fe56643d60a76fc19e5/changelog_test.go

spec.yaml
```diff
@@ -759,3 +762,59 @@ paths:
           description: Error
           schema:
             $ref: "definitions.yaml#/definitions/V4GenericResponse"
+
+  /v4/releases/:
```
xh3b4sd: The name is fine for me. No strong feelings. We should only make sure at some point to align the wording with our internal code in case this will be established.

marians:
Ping @teemow

spec.yaml
puja108: I like the structured approach, which would enable us to automatically generate these. I'm open to think about additional field(s) to allow for more prosaic descriptions of a release to add to user friendliness

marians: So we need a `description` field per component. What if that component hasn't changed since the last release? Make the description `null` comes to my mind.

marians: I think it's semantically much more useful when there is a description per component. The description in the release item serves as the optional additional way to add human-friendly information.

teemow:
LGTM

marians:
@xh3b4sd Is that "make description a list" a blocker for you? Otherwise I'd be happy for an approval stamp.

xh3b4sd:
> Is that "make description a list" a blocker for you?

I would say yes. To me it is unclear how we create this single description string. What we have in the backend is a list of descriptions. How to compute the list of strings into a single one?

marians:
True. I had to look back at our Slack conversation to understand that this really doesn't make sense. Change coming.

marians:
So I changed the example only (TODO: update response schema). See 

https://github.com/giantswarm/api-spec/blob/c6a8c4c19371a85495ffbe7d82285ba88b540f22/spec.yaml#L787-L836

I added the "changelog" key to mimick the versionbundles changelog. This probably should also contain change infos on dependencies.

Dependencies are not exposed on their own so far. Should they? I consider them as something deeper "under the hood" than components.

Note that there is still the main "description" for the release, where I hope we can (but don't have to) publish human-friendly info on a release, especially when it's an important one.

xh3b4sd:
IMO the changelogs are already "human-friendly". We write them manually. I do still not know how we should compute the main description. We can do this somehow though if it is really needed. Lets learn how to do this along the way. 

+1 on not exposing dependencies. 

Being transparent with the changelogs works for me. Does your idea of the changelog description align with the changelog kind we use internally? See https://github.com/giantswarm/versionbundle/blob/efdbdae719c56ab4880f8fe56643d60a76fc19e5/changelog.go#L53.

marians:
@xh3b4sd The changelog entry "description" is a direct copy of the "description" in the versionbundles's changelog entry. Example:

```
{
   "component": "flannel",
   "description": "Flannel version updated.",
  "kind": "changed"
}
```

So far I don't plan to use the "kind" attribute.

spec.yaml
```diff
@@ -759,3 +762,75 @@ paths:
...
+                      "version": "0.9.1"
+                    }
+                  ],
+                  "deprecated": true
+                },
+                {
+                  "version": "2.8.4",
+                  "timestamp": "2017-11-11T12:24:56.59969Z",
+                  "description": "Minor update of calico, fixing a bug.",
+                  "changelog": [
```
xh3b4sd: I noticed this should be plural to be aligned with components.

marians:
Changes with last commit:

- Removed top-level description, since we don't know how to generate it automatically.
- Aligned V4ReleaseListItem definition with the example shown in https://github.com/giantswarm/api-spec/pull/54#issuecomment-342509564

spec.yaml
marians: I wouldn't do that. According to my understanding, a changelog is one changelog, even if it has multiple items in it. Otherwise it should be named "changes" or something like that.

xh3b4sd: Components is a list of things. It is descried in plural form. Changelogs is a ist of things. It is described in singular form. That does not make any sense to me. Do what you think is right. Disagree and commit.

xh3b4sd:
FYI this is implemented and depoyed.

marians:
Merging this. 🎱

PR #59, opened 2017-11-13
Adding release_version to the modifyCluster operation

marians:
Part of https://github.com/giantswarm/giantswarm/issues/2022

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -43,6 +43,9 @@ definitions:
       owner:
         type: string
         description: Name of the organization owning the cluster
+      release_version:
+        type: string
+        description: Release version to use after an upgrade
       workers:
         type: array
         description: Worker node array
--- a/spec.yaml
+++ b/spec.yaml
@@ -338,6 +338,10 @@ paths:
         ownership from one organization to another. The user performing the
         request has to be a member of both organizations.
 
+        - `release_version`: By changing this attribute you can upgrade a
+        cluster to a newer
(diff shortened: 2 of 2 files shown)


definitions.yaml
```diff
@@ -43,6 +43,9 @@ definitions:
       owner:
         type: string
         description: Name of the organization owning the cluster
+      release_version:
+        type: string
+        description: Release version to use after an upgrade
```
xh3b4sd: Full stop is missing IMO.

xh3b4sd: What means after an upgrade? It is the release version used to do the update to. We cannot add a release after the cluster is updated.

marians: The modifyCluster operation (the PATCH request to /v4/clusters/<cluster_id>/) that modifies the `release_version` attribute of the cluster is supposed to initiate a guest cluster upgrade. (We also plan automated updates. Unsure if these use the method described here. However, this is what I think users should use.)

So cluster current has e. g. `release_version=1.2.34`. User does

```
PATCH /v4/clusters/clstr/
{
  "release_version": "1.3.0"
}
```

which starts the cluster upgrade.

marians: I think you and me means the same. "after an upgrade" simply means: once the upgrade is done.

marians:
@xh3b4sd Can you check again if this worth an approval?

xh3b4sd:
FYI this is already implemented. I think we can merge.

PR #61, opened 2017-11-24
Removing limitation regarding scaling on KVM

marians:
For https://github.com/giantswarm/giantswarm/issues/1678

As always, to be merged once the change is deployed.


oponder:
Cool! thanks for making this already.

PR #62, opened 2017-11-27
Removing empty "parameters" key

marians:
Some validator complains here


marians:
Moving on here, as this is blocking me

PR #68, opened 2018-01-11
Remove host and schemes fields

marians:
The host and schemes field, in combination, results in a base URI of `https://api.giantswarm.io/` being communicated in the spec/docs and being hard coded into generated clients. This makes no sense, as we have a specific base URI per installation.


marians:
Re-ping @oponder

Issue #50, opened 2017-10-05
Expose ingress ports in guest cluster details


marians:
Customers ask for details regarding which ports are currently used by a guest cluster's ingress controller. There are two specific ports, one for HTTPS and one for HTTP.

For specs we should have in mind that there might be additional ports to be exposed.

This only affects KVM-based installations currently.

marians:
Done

PR #72, opened 2018-02-28
Spec the ingress ports addition.

oponder:
To make this story a bit more concrete. RFC @marians @xh3b4sd

Already working on the `kubernetesd` part, but wanted to pause a bit and see if we are aligned for the final result of this, in terms of what will actually be added / changed in the API.

https://github.com/giantswarm/api-spec/issues/50

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -147,11 +147,11 @@ definitions:
   V4PortMappingDefinition:
     type: object
     properties:
-      ingressPort:
+      ingress_port:
         description:
           The port on this cluster's ingress.
         type: integer
-      lbPort:
+      load_balancer_port:
         description: |
           The port on the load balancer that is mapped to this cluster's ingress.
         type: integer
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -139,7 +139,7 @@ definitions:
       port_mappings:
         type: array
         description: |
-          Reveals the ports on the load balancer that are mapped to this cluster's ingress
+          Reveals the ports on the host cluster that are mapped to this guest cluster's ingress
           and which protocol that port supports. Only shown and relevant on our on-prem KVM clusters.
         items:
           $ref: '#/definitions/V4PortMappingDefinition'
@@ -147,9 +147,9 @@ definitions:
   V4PortMappingDefinition:
     type: object
     properties:
-      load_balancer_port:
+      port:
         description: |
-          The port on the load balancer that is mapped to this cluster's ingress.
+          The port on the host cluster that will forward traffic to the guest cluster
         type: integer
       protocol:
         description: |
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -136,13 +136,17 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/V4NodeDefinition'
-      port_mappings:
-        type: array
-        description: |
-          Reveals the ports on the host cluster that are mapped to this guest cluster's ingress
-          and which protocol that port supports. Only shown and relevant on our on-prem KVM clusters.
-        items:
-          $ref: '#/definitions/V4PortMappingDefinition'
+      kvm:
+        type: object
+        description: Attributes specific to clusters running on KVM (on-prem) installations.
+        properties:
+          port_mappings:
+            type: array
+            description: |
+              Reveals the ports on the host cluster that are mapped to this guest cluster's ingress
+              and which protocol that port supports. Only shown and relevant on our on-prem KVM clusters.
+            items:
+              $ref: '#/definitions/V4PortMappingDefinition'
 
   V4PortMappingDefinition:
     type: object
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -136,7 +136,29 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/V4NodeDefinition'
+      port_mappings:
+        type: array
+        description: |
+          Reveals the ports on the load balancer that are mapped to this cluster's ingress
+          and which protocol that port supports. Only shown and relevant on our on-prem KVM clusters.
+        items:
+          $ref: '#/definitions/V4PortMappingDefinition'
 
+  V4PortMappingDefinition:
+    type: object
+    properties:
+      ingressPort:
+        description:
+          The port on this cluster's ingress.
+        type: integer
+      lbPort:
+        description: |
+          The port on the load balancer that is mapped to this cluster's ingress.
+        type: integer
+      protocol:
+        description: |
+          The protocol this port mapping is made for.
+        type: string
 
   # Definition of a cluster node
   V4NodeDefinition:
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -147,10 +147,6 @@ definitions:
   V4PortMappingDefinition:
     type: object
     properties:
-      ingress_port:
-        description:
-          The port on this cluster's ingress.
-        type: integer
       load_balancer_port:
         description: |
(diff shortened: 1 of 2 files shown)


oponder:
I'm not sure about the `port_mappings` field name by the way. I chose that because that is kind of how I understand this concept. If there is a more appropriate term, please lets use that.

here is what a `GET /v4/clusters/{id}/` could look like now:

```
              {
                "id": "wqtlq",
                "create_date": "2017-03-03T10:50:45.949270905Z",
                "api_endpoint": "https://api.wqtlq.example.com",
                "name": "Just a Standard Cluster",
                "release_version": "2.5.16",
                "kubernetes_version": "",
                "owner": "acme",
                "workers": [
                  {
                    "memory": {"size_gb": 2},
                    "storage": {"size_gb": 20},
                    "cpu": {"cores": 4},
                    "labels": {
                      "beta.kubernetes.io/arch": "amd64",
                      "beta.kubernetes.io/os": "linux",
                      "ip": "10.3.11.2",
                      "kubernetes.io/hostname": "worker-1.x882ofna.k8s.gigantic.io",
                      "nodetype": "hicpu"
                    }
                  },
                  {
                    "memory": {"size_gb": 8},
                    "storage": {"size_gb": 20},
                    "cpu": {"cores": 2},
                    "labels": {
                      "beta.kubernetes.io/arch": "amd64",
                      "beta.kubernetes.io/os": "linux",
                      "ip": "10.3.62.2",
                      "kubernetes.io/hostname": "worker-2.x882ofna.k8s.gigantic.io",
                      "nodetype": "hiram"
                    }
                  }
                ],
                "port_mappings": [
                  {
                    "ingressPort": 30010,
                    "lbPort": 30020,
                    "protocol": "http"
                  },
                  {
                    "ingressPort": 30011,
                    "lbPort": 30021,
                    "protocol": "https"
                  },
                ]
              }
```

oponder:
> The mapping is from the ingress controller on the host cluster to the service port of the guest cluster pods (which run on the host cluster as well).

> ok why then is it called lbPort?

> Because the `lbPort` is what customers have to configure in their LB on prem.
> Then it goes [LB] -> [host cluster ingress] -> [guest cluster service].

Still doesn't really compute to me :(
It doesn't really line up with the way we've been talking about it in issues so far.

I want to explain this concept well though. 

Is there anyone else that can take a stab at explaining what is actually going on here to me, and what names would make sense to people looking at this? @puja108 perhaps?

oponder:
Like, do my descriptions make sense in definitions.yaml?

definitions.yaml
```diff
@@ -136,7 +136,29 @@ definitions:
...
+        description: |
+          Reveals the ports on the load balancer that are mapped to this cluster's ingress
+          and which protocol that port supports. Only shown and relevant on our on-prem KVM clusters.
+        items:
+          $ref: '#/definitions/V4PortMappingDefinition'
 
+  V4PortMappingDefinition:
+    type: object
+    properties:
+      ingress_port:
```
xh3b4sd: I am not sure this should be exposed. This information is more or less useless because the customer cannot do anything with it AFAIK because this is for the k8s service manifest of the guest cluster which runs in our host cluster. The ports are also always the same. They will always be `30010` for HTTP and `30011` for HTTPS IIRC. See https://github.com/giantswarm/kubernetesd/blob/a790b6be50f18dfe1dd6f7fbd504e8ff2c333fc0/service/creator/ingress.go#L60-L103. 

The actual ingress-controller ports are configured somewhere else AFAIK.

definitions.yaml
```diff
@@ -136,7 +136,29 @@ definitions:
...
+        description:
+          The port on this cluster's ingress.
+        type: integer
+      load_balancer_port:
+        description: |
+          The port on the load balancer that is mapped to this cluster's ingress.
```
xh3b4sd: It maybe helps to use the concepts of customer/company relationship and the concept of our host cluster wording. I am biased but something like this is most clear to me personally. 
 
```
The port on the customer's load balancer that is mapped to the Giant Swarm host cluster ingress-controller.
```

oponder: hmm thx, it gets a bit wordy for me, and doesn't help me understand.

The host cluster shouldn't even have anything to do with anything here. We're looking at details of a guest cluster. 

So this is the port on the load balancer that lets you talk to the ingress of your guest cluster. whether or not the host cluster is in between there is not relevant right?

xh3b4sd: > The host cluster shouldn't even have anything to do with anything here.

This is how it works. I just state facts and we should be as clear as possible to make customers understand what they are doing. IMO transparency is a good thing. I also only tried to help. You can gladly do what you want here. Maybe lets just go with the best we come up and iterate on it. When we notice there is a common misunderstanding over and over again we can adapt. 

> So this is the port on the load balancer that lets you talk to the ingress of your guest cluster. whether or not the host cluster is in between there is not relevant right?

It is relevant to me in order to understand how it works. When you want to hide this to the customer, fine with me. Your sentence I quoted here is quite accurate in summary I guess.

oponder: Thanks for the help! Wasn't rejecting it out right, just trying to get closer to the real understanding, but from the outsiders perspective.

puja108: I think there's a misunderstanding there. The host cluster has quite a lot to do with this as we are talking about onprem here.

In onprem the host cluster ingress controller forwards TCP ports to the guest cluster ingress controller. The guest cluster ingress controller needs 2 of those (http/https). Thus, needs to forward traffic on 2 ports from his (hardware) lb to host cluster IC.

The thing we call lbport is actually a port on the host cluster (behind which the host cluster ingress controller sits). So it is not the port of the customer's load balancer. On the customer's load balancer that is usually sth like IP:443 and IP:80 (where IP is the dedicated IP for a single guest cluster).

So a typical example for a guest cluster https request would look like following:

optional outside DNS -> https://ingress.clusterid.basedomain (DNS)-> IP:443 (on hw lb) -> 30021 (on host cluster) -> host cluster IC TCP service for cluster x -> 30011 (on guest cluster ) -> guest cluster IC https -> (service (only for endpoint lookup in IC -> ) pod -> container

definitions.yaml
puja108: true, this is internal and always the same, also the customer can look this up inside their cluster, as this is basically their IC config

definitions.yaml
```diff
@@ -136,7 +136,29 @@ definitions:
...
+          and which protocol that port supports. Only shown and relevant on our on-prem KVM clusters.
+        items:
+          $ref: '#/definitions/V4PortMappingDefinition'
 
+  V4PortMappingDefinition:
+    type: object
+    properties:
+      ingress_port:
+        description:
+          The port on this cluster's ingress.
```
puja108: external port of the cluster's ingress controller

definitions.yaml
puja108: Maybe `This is the port on the host cluster ingress controller that needs to be set up to receive traffic for this cluster.`

puja108: or even minus the `to be set up` it needs to receive traffic, how you solve that, is up to you, and you most probably know if you are looking for this.

oponder:
How about "The port on the host cluster that will forward traffic to the guest cluster" ?

oponder:
Anyone still opposed to anything in here? Can I get an approve?

oponder:
@marians @puja108 @xh3b4sd This spec is finally implemented. Though while implementing I ended up scoping the `port_mappings` under a `kvm` object since this is a `kvm` specific piece of info.

PR #75, opened 2018-03-07
adds release version to cluster lister response

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2760.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -506,7 +506,7 @@ definitions:
         description: Name of the organization owning the cluster
       release:
         type: string
-        description: The semantic version number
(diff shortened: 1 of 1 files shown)


definitions.yaml
```diff
@@ -504,6 +504,9 @@ definitions:
       owner:
         type: string
         description: Name of the organization owning the cluster
+      release:
+        type: string
+        description: The semantic version number
```
oponder: can we add "of this cluster" to the end of the sentence?

xh3b4sd: Done.

PR #80, opened 2018-03-28
removes kubernetes version from cluster creator

xh3b4sd:
Follow up of https://github.com/giantswarm/cluster-service/pull/278.


xh3b4sd:
Then it stays in the docs but does not have any effect. I would consider this a breaking change already. How to deal with this then? The property is already being ignored. I just want to cleanup.

marians:
> Then it stays in the docs but does not have any effect.

Right, that would be fine with me.

> I would consider this a breaking change already.

With "this", do you think the fact that the attribute has no effect is a breaking change?

> How to deal with this then? The property is already being ignored. I just want to cleanup.

I think we have to live with it until there is a v5.

xh3b4sd:
> With "this", do you think the fact that the attribute has no effect is a breaking change?

Yes, the semantics changed already. Keeping the property which has no impact anymore has no benefit. I understand it to keep the docs for the deprecation notice but I think we can remove the property in our code. When people still send it the API should actually just ignore it.

marians:
> When people still send it the API should actually just ignore it.

Correct.

PR #81, opened 2018-04-10
Add addCredentials operation

marians:
Towards
- https://github.com/giantswarm/giantswarm/issues/2944
- https://github.com/giantswarm/giantswarm/issues/2652 (epic)

This PR adds the spec for a `addCredentials` operation that allows to set the provider credentials for an organization.

### Preview

For an AWS installation and an organization named `acme`, this is how credentials could be specified:

```nohighlight
curl -X POST -H "Authorization: giantswarm $TOKEN" $API_ENDPOINT/v4/organizations/acme/credentials/ -d '
{
  "provider": "aws",
  "aws": {
    "roles": {
      "admin": "arn:aws:iam::123456789012:role/GiantSwarmAdmin",
      "awsoperator": "arn:aws:iam::123456789012:role/GiantSwarmAWSOperator"
    }
  }
}'
```

Response:

```nohighlight
HTTP/1.1 201 Created
Location: /v4/organizations/acme/credentials/abc123/
Content-type: application/json

{
  "code": "RESOURCE_CREATED",
  "message": "A new set of credentials has been created with ID 'abc123'"
}
```

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -581,7 +581,7 @@ definitions:
         type: string
       aws:
         type: object
-        description: Credentials specific to an AWS
+        description: Credentials specific to an AWS account
         required:
           - account_id
           - access_key_id
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -568,3 +568,31 @@ definitions:
       expiry:
         type: string
         description: The date and time when this account will expire
+
+  V4AddCredentialsRequest:
+    type: object
+    required:
+      - provider
+    description: Request model for adding a set of credentials
+    properties:
+      provider:
+        type: string
+      description:
+        type: string
+      aws:
+        type: object
+        description: Credentials specific to an AWS
+        required:
+          - account_id
+          - access_key_id
+          - secret_access_key
+        properties:
+          account_id:
+            type: string
+            description: AWS account ID (usually a numeric string)
+          access_key_id:
+            type: string
+            description: AWS access key ID, base46-encoded
+          secret_access_key:
+            type: string
+            description: AWS secret access key ID, base46-encoded
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -577,22 +577,21 @@ definitions:
     properties:
       provider:
         type: string
-      description:
-        type: string
       aws:
-        type: object
+        type: array
         description: Credentials specific to an AWS account
-        required:
-          - account_id
-          - access_key_id
-          - secret_access_key
-        properties:
-          account_id:
-            type: string
-            description: AWS account ID (usually a numeric string)
-          access_key_id:
-            type: string
-            description: AWS access key ID, base46-encoded
-          secret_access_key:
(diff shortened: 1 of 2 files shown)


definitions.yaml
```diff
@@ -568,3 +568,31 @@ definitions:
...
+      - provider
+    description: Request model for adding a set of credentials
+    properties:
+      provider:
+        type: string
+      description:
+        type: string
+      aws:
+        type: object
+        description: Credentials specific to an AWS
```
TheoBrigitte: to an AWS ... ? account?

definitions.yaml
```diff
@@ -568,3 +568,31 @@ definitions:
...
         description: The date and time when this account will expire
+
+  V4AddCredentialsRequest:
+    type: object
+    required:
+      - provider
+    description: Request model for adding a set of credentials
+    properties:
+      provider:
+        type: string
```
xh3b4sd: Why do we need the provider when we have the `aws` block?

spec.yaml
```diff
@@ -997,6 +997,58 @@ paths:
...
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v4/organizations/{organization_id}/credentials/:
+    post:
+      operationId: addCredentials
+      tags:
+        - organizations
+      summary: Set credentials
+      description: |
+        Add a set of credentials to the organization allowing to create and
```
rossf7: "allowing to" sounds a bit strange to me. "allowing the creation and operation of clusters" is better IMO.

oponder:
> "Credentials in an organization are immutable. An organization can only have one set of credentials."


Have we addressed how we'll edit credentials then if the user makes a typo?

marians:
> Have we addressed how we'll edit credentials then if the user makes a typo?

This is beyond the API in the first iteration. The workaround will require creating a new organization.

JosephSalisbury:
@marians do you have some thoughts on error responses?

- if the organization does not exist?
- if the provider key != aws?
- if the aws key is empty?
- if the roles list is empty?
- if the roles list does not have two entries?
- if the arn key is not valid?

kopiczko:
Also Secret format will be close to that: https://github.com/giantswarm/giantswarm/issues/3070#issuecomment-385345083

Reason: simplified validation. Something to think about here.

spec.yaml
```diff
@@ -997,6 +997,78 @@ paths:
...
+        ```json
+        {
+          "provider": "aws",
+          "aws": {
+            "roles": {
+              "admin": "arn:aws:iam::123456789012:role/GiantSwarmAdmin",
+              "awsoperator": "arn:aws:iam::123456789012:role/GiantSwarmAWSOperator"
+            }
+          }
+        }
```
kopiczko: <3

marians:
Merging as this is implemented. Thanks folks!

PR #82, opened 2018-05-11
Add a spec for GET /v4/users/{email}

oponder:
This adds a quick spec for /v4/users/{email}

(For which I have an implementation in progress already. I made it as part of the JWT token exploration.)

This endpoint will return info about a specific user, or about the currently authenticated user if thet special value of `me` is used for the email parameter.

RFR


marians:
Thanks for taking this.

I am not in favour of having (copying) the `/v4/users/me/` exception, for consistency reasons obviously. Please let's define a separate endpoint for that, e. g. `/v4/myself/`, `/v4/currentuser/` or `/v4/me/`.

oponder:
Thanks for looking!

I don't agree with having a separate endpoint for this exception though.

`me` isn't a valid email, so we can be sure there won't be any collisions there.

`/v4/users/me/` does represent a valid 'resource` though. Granted, it's a resource that changes based on who is talking to it, but that would also be the case for `/v4/myself/`.

What are we trying to be consistent with by creating a new endpoint? The collection of users is at `/v4/users/`, and a specific user is at `/v4/users/{specificuser}/` , that is pretty conventional and normal right?

marians:
My gut tells me that we will regret that exception, as it can mean extra treatment in many places.

Here is another argument: The behaviour of an endpoint should be easily explainable. With `/v4/users/{email}/`, that is the case. A dedicated endpoint to return info on the current user would also be easily explainable. A combination of both is less so.

Also I don't see a good argument for combining the two functions.

oponder:
Ok, how about the naming of the new endpoint though? 

How do you feel about: `/v4/user/` ?

oponder:
Can I get a approve? Removed the part that can be implemented independently.

oponder:
@marians I think you re-approved here when you meant to approve in #83 ?

PR #83, opened 2018-05-11
Add a spec for fetching info about a specific user. (/v4/users/{email}/)

oponder:
Specs the `/v4/users/{email}/` endpoint.

Separated from: https://github.com/giantswarm/api-spec/pull/82
Part of: Part of: https://github.com/giantswarm/giantswarm/issues/2955

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -201,6 +201,48 @@ paths:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/users/{email}/:
+    get:
+      operationId: getUser
+      parameters:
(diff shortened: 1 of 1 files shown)


spec.yaml
```diff
@@ -170,6 +170,48 @@ paths:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/users/{email}/:
+    get:
+      operationId: getUser
```
marians: `getUser` conflicts with the new operation defined in #82

oponder: Ok, adjusted in #82, made the operationId there be `getCurrentUser`

PR #84, opened 2018-05-22
Include ReDoc JS file in docker image

marians:
This avoids relying on ReDoc being available at runtime and also reduced the amount of third party tracking. See https://github.com/giantswarm/giantswarm/issues/3232


marians:
@oponder Tiny review possible?

PR #89, opened 2018-06-26
Fix type of storage.size_gb and clarify description

marians:
For https://github.com/giantswarm/giantswarm/issues/3449

This fixes a problem with a wrong type definition (int vs. number) and also makes it more explicit in the description that here, both an integer and a floating point value can occur.


marians:
Thanks @puja108 for reacting to sig-ux requests :)

PR #91, opened 2018-07-03
Add request examples for API docs, improve examples usability

marians:
Solves https://github.com/giantswarm/giantswarm/issues/3619

This should add request examples to all operations that have a request body.

Note:

The `x-examples` key is a vendor extension introduced by ReDoc, the tool we use for rendering our docs. Swagger/OpenAPI 2.0 does not yet support request examples. This solves a problem we had with request examples being auto-generated and pretty much meaningless.

Also made the displayed `size_gb` value in an example a float instead of an integer, to show that floating point values might occur.

In addition, renamed tag `auth-tokens` to `auth tokens`, to align with `key pairs`.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -599,8 +599,8 @@ paths:
                 "owner": "acme",
                 "workers": [
                   {
-                    "memory": {"size_gb": 2},
-                    "storage": {"size_gb": 20},
+                    "memory": {"size_gb": 2.0},
+                    "storage": {"size_gb": 20.0},
                     "cpu": {"cores": 4},
                     "labels": {
                       "beta.kubernetes.io/arch": "amd64",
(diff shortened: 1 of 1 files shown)


spec.yaml
```diff
@@ -921,6 +973,15 @@ paths:
           required: true
           schema:
             $ref: "./definitions.yaml#/definitions/V4Organization"
+          x-examples:
```
oponder: I just double checked since I wasn't sure if this would work,
but the `api` doesn't support creating and adding users in one API call currently.

marians: I don't understand the point. Is that related to the example?

oponder: Yeah the example block below, if I am right, refers to the `addOrganization` operationId.

it shows the `members` field in the example, populated with some sample members.

If you actually try to create an organisation like this, these members will not be added to the organization.

marians: OK. Created an issue: https://github.com/giantswarm/giantswarm/issues/3620

marians:
Thanks! Not aware of a way to expand by default.

PR #93, opened 2018-07-04
Document cn_prefix charset in API spec

marians:
For https://github.com/giantswarm/api/issues/475

It seems like vault's PKI implementation requires the entire CN to be a valid domain name (in terms of the character set used). I tested this myself, as I couldn't get a source for the actual validation in place.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -256,7 +256,7 @@ definitions:
         description: Expiration time (from creation) in hours
       cn_prefix:
         type: string
-        description: The common name prefix of the certificate subject.
+        description: The common name prefix of the certificate subject. This only allows characters that are usable in domain names (`a-z`, `0-9`, and `.-`, where `.-` must not occur at the start nor end).
       certificate_organizations:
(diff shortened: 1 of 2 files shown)


definitions.yaml
```diff
@@ -256,7 +256,7 @@ definitions:
         description: Expiration time (from creation) in hours
       cn_prefix:
         type: string
-        description: The common name prefix of the certificate subject.
+        description: The common name prefix of the certificate subject. This only allows characters that are usable in domain names (`a-z`, `0-9`, and `.-`, where `.-` must not occur at the start nor end).
       certificate_organizations:
```
pipo02mix: Nitpick 
I think it is `or` instead of `nor` at the end of the phrase
`at the start or end`

JosephSalisbury: 🇬🇧 driveby:

```
where `.-` must not occur at either the start or the end).
```

would be how I would right it. `nor` is super old school.

PR #96, opened 2018-07-10
Add optional request headers

marians:
Related to https://github.com/giantswarm/gsctl/issues/251

As we are abandoning gsclientgen's custom spec and using the official api-spec instead, I found that we are missing the specification for these optional headers.


marians:
I don't see a way to simplify this :(

PR #97, opened 2018-07-12
Add 401 responses everywhere

marians:
Related to https://github.com/giantswarm/gsctl/pull/251

This PR ensures that there is a 401 response defined for every operation. It uses referenes to one generic example for that.

Rationale is that this allows the new client to operate in a consistent way accross all operations. The code generated by go-swagger is different, depending on whether a response is defined or not. To have it consistent we could either have no specific responses anywhere and use the "default" for anything other than 200 (which is bad from a documentation point of view) or have every specific response specified.

Expect more of these PRs the near future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+PWD := $(shell pwd)
 PHONY: lint spec-tmp.yaml
 
 SWAGGER_VERSION=2.2.3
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,7 +16,12 @@ spec-tmp.yaml:
 
 # YAML linting
 lint: spec-tmp.yaml
-	yamllint -c ./yamllint/config.yaml ./spec.yaml ./definitions.yaml ./parameters.yaml ./responses.yaml
+	docker run --rm -ti \
+		-v $(PWD):/workdir \
+		-w /workdir \
+		giantswarm/yamllint \
+		-c ./yamllint/config.yaml \
+		./spec.yaml ./definitions.yaml ./parameters.yaml ./responses.yaml
```
marians: This change is unrelated. It allows to run the linting without having `yamllint` available.

PR #100, opened 2018-07-17
Add patch user endpoint

oponder:
This PR adds a spec for PATCH to our user resource endpoint.

~PATCH `/v4/user/`:~
~Allows someone to modify their own e-mail address.~

PATCH `/v4/users/:email/`:
Allows to modify a given user's e-mail address and/or their expiry date. Only admins are able to edit accounts that don't belong to them, as well as the expiry date.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -569,14 +569,6 @@ definitions:
         type: string
         description: The date and time when this account will expire
 
-  V4ModifyCurrentUserRequest:
-    type: object
-    description: Request model for modifying your own profile
-    properties:
-      email:
-        type: string
-        description: Set the new email address you would like to have.
-
   V4ModifyUserRequest:
     type: object
     description: Request model for modifying a specific user as an admin
@@ -586,7 +578,7 @@ definitions:
         description: Set the new email address you would like the user to have.
       expiry:
         type: string
-        description: Set the new expiry date you would like the user to have.
+        description: Set the new expiry date you would like the user to have. (Only editable by admins)
 
   V4AddCredentialsRequest:
     type: object
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -569,6 +569,25 @@ definitions:
         type: string
         description: The date and time when this account will expire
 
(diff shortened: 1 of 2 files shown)


oponder:
**Topic: Email verification**
When a user changes their own email address, we should actually do a email verification like we do during invitation to ensure the user actually owns the email address that they are trying to set.

This would probably change the process so that PATCH `/v4/user/` would return a `202 accepted` code. 

Do you think we should try to implement this feature now? Or stick with the status quo.

definitions.yaml
```diff
@@ -569,6 +569,17 @@ definitions:
         type: string
         description: The date and time when this account will expire
 
+  V4ModifyUserRequest:
+    type: object
+    description: Request model for modifying a specific user as an admin
```
marians: Please remove ` as an admin`

definitions.yaml
```diff
@@ -569,6 +569,17 @@ definitions:
...
+    properties:
+      email:
+        type: string
+        description: Set the new email address you would like the user to have.
+      expiry:
+        type: string
+        description: Set the new expiry date you would like the user to have. (Only editable by admins)
```
marians: See above ("Set the")

spec.yaml
```diff
@@ -382,6 +382,45 @@ paths:
...
+                "expiry": "2037-01-15T00:00:00Z"
+              }
+      responses:
+        "200":
+          description: Success
+          schema:
+            $ref: "./definitions.yaml#/definitions/V4UserListItem"
+          examples:
+            application/json:
+              {"email": "edited-email@example.com", "created": "2017-01-15T12:00:00Z", "expiry": "2037-01-15T00:00:00Z"}
```
marians: Please indent the JSON for improved readability

oponder:
> Changing the password should not happen via this endpoint?

I'm open to that. But see discussion here: https://gigantic.slack.com/archives/C13MG9F0W/p1527521425000092

oponder:
@marians: Ok! Changes applied!

PR #101, opened 2018-07-17
Add POST /v4/users/:email/password/ endpoint

oponder:
This PR adds the endpoint that allows users and admins to modify their password.


definitions.yaml
```diff
@@ -569,6 +569,19 @@ definitions:
         type: string
         description: The date and time when this account will expire
 
+  V4ModifyUserPassword:
+    type: object
+    required:
+      - new_password_base64
```
oponder: Not sure about this required section.

`new_password_base64` is required for anyone accessing this endpoint, but `current_password_base64` only applies to users editing their own password and not admins.

pipo02mix: Why not make this same for both? Now admin will login via SSO anyway

oponder: this is about the case where an admin wants to edit the password of someone else. 

so he doesn't know that persons 'current_password'

marians: I'd like to rename `V4ModifyUserPassword` to `V4ModifyUserPasswordRequest`.

We have only paid a bit of attention to this in the past. But as this can affect the APIs of generated clients, I'd like us to be consistent if we can.

oponder: Thanks for noticing! Sorry I'm a bit inconsistent in noticing these details.

marians: That's what reviews are for :)

PR #103, opened 2018-08-16
Add spec for inviting users.

oponder:
This adds the spec for inviting users (POST /v4/invitations/)
As well as a way of inspecting the invitations made so far (GET /v4/invitations/)

This goes towards: https://github.com/giantswarm/giantswarm/issues/3831 and https://github.com/giantswarm/giantswarm/issues/3965

Up next would be a way to actually accept the invitation (POST /v4/invitations/:token/)
And a way to validate the invitation token so that clients can provide some feedback early in the process (GET /v4/invitations/:token/)

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -683,12 +683,12 @@ definitions:
       created:
         type: string
         description: Date/Time that the invitation was created
-      expiry:
-        type: string
-        description: Date/Time that the invitation will expire
       email:
         type: string
         description: Email of the invitee
+      expiry:
+        type: string
+        description: Date/Time that the invitation will expire
       invited_by:
         type: string
         description: Email of the person that created the invitation
@@ -717,24 +717,3 @@ definitions:
       send_email:
         type: boolean
         description: Whether or not to send the invitee an email. Defaults to true
-
-  V4CreateInvitationResponse:
-    type: object
-    properties:
-      created:
-        type: string
-        description: Date/Time that the invitation was created
-      email:
-        type: string
-        description: Email of the invitee
-      invitation_accept_link:
-        type: string
-        description: A link that can be used to accept the invitation.
-      invited_by:
-        type: string
-        description: Email of the person that created the invitation
-      organizations:
-        type: array
-        description: List of organizations that the invitee will be added to
-        items:
-          type: string
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -683,6 +683,9 @@ definitions:
       created:
         type: string
         description: Date/Time that the invitation was created
+      expiry:
+        type: string
+        description: Date/Time that the invitation will expire
       email:
         type: string
         description: Email of the invitee
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -747,4 +747,3 @@ definitions:
       status:
         type: string
         description: The status of the invitation. Either 'pending', 'accepted', or 'expired'
-
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -680,9 +680,6 @@ definitions:
       accepted:
         type: string
         description: Date/Time that the invitation was accepted
-      account_expiry_days:
-        type: integer
-        description: The number of days after which the account should expire
       created:
         type: string
         description: Date/Time that the invitation was created
@@ -706,9 +703,6 @@ definitions:
     required:
       - email
     properties:
-      account_expiry_days:
-        type: integer
-        description: The number of days after which the account should expire. Defaults to '0', which means never expire
       email:
         type: string
         description: The email of the person you want to invite
@@ -724,9 +718,6 @@ definitions:
   V4CreateInvitationResponse:
     type: object
     properties:
-      account_expiry_days:
-        type: integer
-        description: The number of days after which the account should expire
       created:
         type: string
         description: Date/Time that the invitation was created
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -666,3 +666,85 @@ definitions:
       auth_token:
         type: string
         description: The newly created API token
+
+  # The list of all invitations.
+  V4InvitationsResponse:
+    type: array
+    description: List of members that belong to this organization
+    items:
+      $ref: '#/definitions/V4Invitation'
+
+  V4Invitation:
+    type: object
+    properties:
+      accepted:
+        type: string
+        description: Date/Time that the invitation was accepted
+      account_expiry_days:
+        type: integer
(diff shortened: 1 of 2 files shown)


definitions.yaml
```diff
@@ -666,3 +666,84 @@ definitions:
       auth_token:
         type: string
         description: The newly created API token
+
+  # The list of all invitations.
+  V4InvitationsResponse:
+    type: array
+    description: List of members that belong to this organization
```
puja108: this description sounds like it will be all members of the org, shouldn't this a list of all invites (no matter which status)?

definitions.yaml
```diff
@@ -666,3 +666,84 @@ definitions:
...
+    items:
+      $ref: '#/definitions/V4Invitation'
+
+  V4Invitation:
+    type: object
+    properties:
+      accepted:
```
puja108: this is optional right? if it's pending for example

definitions.yaml
puja108: I know that is similar to all members, but all members sounds to me like excluding pending or expired invites

definitions.yaml
```diff
@@ -666,3 +666,84 @@ definitions:
...
+        type: string
+        description: Email of the person that created the invitation
+      organizations:
+        type: array
+        description: List of organizations that the invitee will be added to
+        items:
+          type: string
+      status:
+        type: string
+        description: The status of the invitation. Either 'pending', 'accepted', or 'expired'
```
puja108: could we have revoked? or is invitation expiry short enough to not need revocation of invites?

definitions.yaml
```diff
@@ -666,3 +666,84 @@ definitions:
...
+
+  V4CreateInvitationResponse:
+    type: object
+    properties:
+      account_expiry_days:
+        type: integer
+        description: The number of days after which the account should expire
+      created:
+        type: string
+        description: Date/Time that the invitation was accepted
```
puja108: that does not make sense, the creation event implies that at that moment the invitation cannot have been accepted. should this be rather the creation timestamp of the invite itself (so we know when the invite expires)?

definitions.yaml
```diff
@@ -666,3 +666,84 @@ definitions:
...
+        type: string
+        description: Email of the person that created the invitation
+      organizations:
+        type: array
+        description: List of organizations that the invitee will be added to
+        items:
+          type: string
+      status:
+        type: string
+        description: The status of the invitation. Either 'pending', 'accepted', or 'expired'
```
puja108: on creation this can only be pending and thus is kind of obsolete IMO

spec.yaml
```diff
@@ -165,6 +165,111 @@ paths:
...
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v4/invitations/:
+    get:
+      operationId: getInvitations
+      tags:
+        - invitations
+      summary: Get invitations
+      description: |
+        Returns a list of invitations that are either pending or have been accepted.
```
puja108: not the expired ones?

spec.yaml
```diff
@@ -165,6 +165,111 @@ paths:
...
+      operationId: createInvitation
+      tags:
+        - invitations
+      summary: Create invitation
+      description: |
+        Use this endpoint to invite someone to your installation. The response
+        will include the invitation_accept_link, which will never be shown again.
+        The link is also not stored and should only be shared with the intended recipient.
+
+        The recipient will be sent a email with a link they can click to accept
```
puja108: `an email`

definitions.yaml
oponder: Currently no way to revoke an invite. We could create a DELETE endpoint on an invite, we would delete the object then, and there wouldn't be a `revoked` state left. 

I think this feature doesn't need that much of a paper trial. but that is up to debate.

definitions.yaml
oponder: Yeah I think I have some bad copy paste here.

definitions.yaml
oponder: Yes, it can be blank. But this is part of a response, so there is no requried or optional specification here.

puja108: ok, was just not sure if there's a way to mention sth about optionality in the spec as otherwise an api user might expect there to be such a field coming back

definitions.yaml
puja108: as long as we have sth like 72h expiration a delete/revoke is not needed

marians:
Regarding account expiry: We discusses that feature internally several times and there seems to be no need for it really any more. So I would propose to not expose it in the API and instead set all expiry to never, so we can get rid of it eventually.

definitions.yaml
oponder: My intention here is that the field will come back, but it'll be empty if the invitation has not yet accepted.

marians:
BTW, I am still reviewing and have more questions

oponder:
Puja mentioned that the response for creating a invitation should not contain the `accepted` key, nor the `status` key, since we know on creation that it is pending, and there would not be an `accepted` date.

However I think we should return the same object we return in the listing of invitations. So I will adjust the spec for that. Even if the the information is obvious, I think the pattern we follow is to return the full 'domain object' after you create it.

definitions.yaml
oponder: > Regarding returning accepted as an empty string when not accepted yet: I consider that a bad practice. I would leave it out (or second best: make it have null as a value) in that case.

I disagree here. Isn't it better to have a single type possible for that field? instead of it being a `null` or `string` ?

definitions.yaml
```diff
@@ -666,3 +666,72 @@ definitions:
...
+  # The list of all invitations.
+  V4InvitationsResponse:
+    type: array
+    description: List of all invitations in the system
+    items:
+      $ref: '#/definitions/V4Invitation'
+
+  V4Invitation:
+    type: object
+    properties:
```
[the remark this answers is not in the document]
oponder: Ok done! added a `expiry` field.

I'm of the opinion we should always return it though, not only when it is `pending` or `expired`

definitions.yaml
marians: This is very much a Go-minded argument. `null` as a special value for an attribute that would otherwise carry a string/a number/an array/an object is pretty common in JSON APIs and even Go clients should be able to deal with this. That said, only returning an attribute like `accepted` when it actually has a meaningful value would make this even simpler.

oponder:
Closing, quite old. Can revisit if it becomes a prio again.

PR #105, opened 2018-08-21
Expose BYOC credential details

marians:
For https://github.com/giantswarm/giantswarm/issues/3987

This PR adds functions to fetch details about the credentials in use for a cluster in AWS or Azure.

- `getCluster` (`GET /v4/clusters/{cluster_id}/`) response now also contains the `credentials_id` key
- new operation `getCredentials` (`GET /v4/organizations/{organization_id}/credentials/`) to fetch credentials of an organization
- new operation `getCredential` (`GET /v4/organizations/{organization_id}/credentials/{credentials_id}/`) to fetch credentials

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ validate: spec-tmp.yaml lint
 	# Useful to ensure gsclientgen client generation works
 	@echo "Validating with go-swagger"
 	docker run --rm -it \
-	  -v ${PWD}:/workdir \
+		-v ${PWD}:/workdir \
 		-w /workdir \
 		quay.io/goswagger/swagger:0.16.0 \
 			validate spec-tmp.yaml
--- a/Makefile
+++ b/Makefile
@@ -25,20 +25,21 @@ lint: spec-tmp.yaml
 
 # Validate the swagger/OAI spec
 validate: spec-tmp.yaml lint
+	
+	# Useful to ensure gsclientgen client generation works
+	@echo "Validating with go-swagger"
+	docker run --rm -it \
+	  -v ${PWD}:/workdir \
+		-w /workdir \
+		quay.io/goswagger/swagger:0.16.0 \
+			validate spec-tmp.yaml
+
 	@echo "Generating swagger representation"
 	docker run --rm -it \
 		-v $(shell pwd):/code/yaml \
 		jimschubert/swagger-codegen-cli:${SWAGGER_VERSION} generate \
 		--input-spec /code/yaml/spec-tmp.yaml \
 		--lang swagger --output /tmp/
-
-	@echo "Generating static HTML documentation"
-	docker run --rm -it \
-		-v $(shell pwd):/code/yaml \
-		-v $(shell pwd)/html:/code/html \
-		jimschubert/swagger-codegen-cli:${SWAGGER_VERSION} generate \
-		--input-spec /code/yaml/spec-tmp.yaml \
-		--lang html --output /code/html
 
 	@echo "Generating JavaScript client code for test purposes"
 	docker run --rm -it \
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -136,6 +136,9 @@ definitions:
       name:
         type: string
         description: Cluster name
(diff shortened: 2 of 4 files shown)


definitions.yaml
```diff
@@ -648,6 +651,57 @@ definitions:
...
+          roles:
+            type: object
+            description: IAM roles to assume by certain entities
+            properties:
+              admin:
+                type: string
+                description: ARN of the IAM role Giant Swarm support staff will use
+              awsoperator:
+                type: string
+                description: ARN of the IAM role assumed by the software operating clusters
```
JosephSalisbury: `ARN of the IAM role assumed by the software operating the clusters` reads a bit easier for me

spec.yaml
```diff
@@ -1246,6 +1247,45 @@ paths:
...
+        "200":
+          description: Credentials
+          schema:
+            $ref: './definitions.yaml#/definitions/V4GetCredentialsResponse'
+          examples:
+            application/json:
+              [
+                {
+                  "id": "a1b2c3",
+                  "provider": "azure",
```
JosephSalisbury: can we have an example response for aws, too?

spec.yaml
```diff
@@ -1324,25 +1364,66 @@ paths:
...
+      responses:
+        "200":
+          description: Credentials details
+          schema:
+            $ref: "./definitions.yaml#/definitions/V4GetCredentialResponse"
+          examples:
+            application/json:
+              {
+                "id": "a1b2c3",
+                "provider": "azure",
```
JosephSalisbury: and here, aws would be useful as well

spec.yaml
```diff
@@ -1324,25 +1364,66 @@ paths:
...
+                "azure": {
+                  "credential": {
+                    "client_id": "c93bf55e-5bf7-4966-ad2b-e6f6e7721d50",
+                    "subscription_id": "b388b7c7-4479-4040-9ac5-1e13edd6b1cd",
+                    "tenant_id": "3dd2e94a-92ba-434c-99be-32bb65864a99"
+                  }
+                }
+              }
```
kopiczko: Github didn't add my comment to the approval:

We agreed we use singular form of `credential` but I see a lot of `credentials` here. Otherwise LGTM.

spec.yaml
marians: We can have only one example response (per content type, to be exact). Unfortunately. But I can add eamples to the description text.

definitions.yaml
marians: changed

Makefile
```diff
@@ -25,20 +25,21 @@ lint: spec-tmp.yaml
...
-	@echo "Generating swagger representation"
+	
+	# Useful to ensure gsclientgen client generation works
+	@echo "Validating with go-swagger"
 	docker run --rm -it \
-		-v $(shell pwd):/code/yaml \
-		jimschubert/swagger-codegen-cli:${SWAGGER_VERSION} generate \
-		--input-spec /code/yaml/spec-tmp.yaml \
-		--lang swagger --output /tmp/
+	  -v ${PWD}:/workdir \
```
TheoBrigitte: Nit: indentation

spec.yaml
marians: Added to the description

spec.yaml
marians: Added to description

marians:
Hey @giantswarm/team-magic , I already asked in https://github.com/giantswarm/giantswarm/issues/3987#issuecomment-414589519 but I don't feel confident at all about my suggestion.

Based on language, `credentials_id` might sound right, but in terms of API jargon consistency, I feel that as it's about a single object, it should be `credential_id`.

TheoBrigitte:
> Based on language, credentials_id might sound right, but in terms of API jargon consistency, I feel that as it's about a single object, it should be credential_id.

`credential_id` feels right to me, so no confusion between singular or plural.
In case of multiple objects I would instead expect something like `credential_ids` where a list of id would be passed.

PR #107, opened 2018-08-31
Render HTML statically using redoc-cli

marians:
For https://github.com/giantswarm/giantswarm/issues/3626

This replaces the current single-page JavaScript application rendering our API docs based on the swaggger YAML with a HTML + minimal JavaScript version.

The benefits are on the SEO and loading speed side, as well as in usability for slower devices. However the final result isn't as polished as the one we currently have. I am also worried that the template or CSS used diverges from ReDoc's master and will lag behind when it comes to improvements.

See for yourself by running

```
make run-server
```

and then open http://localhost:8080/

Preview

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,18 @@
-FROM node:10 as builder
+FROM node:8-slim as builder
 
-WORKDIR /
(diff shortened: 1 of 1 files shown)


marians:
Not pursuing this due to the lower quality result

PR #108, opened 2018-09-04
add docker volume size to node definition

teemow:
this makes the size of the docker volume that is attached to each node.
docker uses this volume to fetch images and depending on the size of the
images being used in the cluster the users would like to configure the
size of this volume.

disclaimer: the data structure of nodes/workers is messed up already.
there are differences between aws/azure and kvm whether the nodes can be
configured individually or not. this is a mess and this addition will
not aim to fix this. this will be changed with the work on node pools.

PR #110, opened 2018-09-05
add kubelet volume size to node definition

teemow:
this makes the size of the kubelet volume that is attached to each node
configurable.

the kubelet uses this volume to store its files and to mount volumes.

disclaimer: the data structure of nodes/workers is messed up already.
there are differences between aws/azure and kvm whether the nodes can be
configured individually or not. this is a mess and this addition will
not aim to fix this. this will be changed with the work on node pools.

built on top of: https://github.com/giantswarm/api-spec/pull/108


definitions.yaml
```diff
@@ -76,6 +76,23 @@ definitions:
                       default:
                         description: Default size of the docker volume, if not specified
                         type: number
+              kubelet:
+                description: Information about the kubelet volume attached to the worker node. This volume is used to store files for the kubelet (volume mounts etc)
```
marians: Please append a period to `etc` and also to the end of the sentence.

marians: How about mentioning that this is also used by `EmptyDir` type volumes?

definitions.yaml
```diff
@@ -241,6 +258,18 @@ definitions:
                 description: |
                   Size of the docker volume attached to the node in GB. Must be
                   the same for all worker nodes of a cluster.
+          kubelet:
+            description: |
+              Volume for the kubelet. Usually mounted at `/var/lib/kubelet`
```
marians: Please append a period after `/var/lib/kubelet`

definitions.yaml
teemow: added. thanks!

PR #112, opened 2018-10-11
adds idea about default provider worker count

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/4115.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -97,7 +97,7 @@ paths:
           },
           "workers": {
             "count_per_cluster": {
-              "max": 20,
+              "max": -1,
               "default": 3
             },
             "instance_type": {
--- a/spec.yaml
+++ b/spec.yaml
@@ -97,7 +97,7 @@ paths:
           },
           "workers": {
             "count_per_cluster": {
-              "max": -1,
+              "max": null,
               "default": 3
             },
             "instance_type": {
@@ -147,7 +147,7 @@ paths:
                 },
                 "workers": {
                   "count_per_cluster": {
-                    "max": 20,
+                    "max": null,
                     "default": 3
                   },
                   "instance_type": {
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -17,55 +17,55 @@ definitions:
     type: object
     properties:
       general:
-        description: General information.
+        description: General information
         type: object
         properties:
           installation_name:
-            description: Unique name of the installation.
+            description: Unique name of the installation
             type: string
           provider:
             description: The technical provider used in this installation. Either "kvm", "aws", or "azure".
             type: string
           datacenter:
-            description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1".
+            description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
       workers:
-        description: Information related to worker nodes.
+        description: Information related to worker nodes
         type: object
         properties:
           count_per_cluster:
-            description: Number of workers per cluster.
+            description: Number of workers per cluster
             type: object
             properties:
               max:
                 description: Maximum number of worker a cluster can have. Might be null when unknown.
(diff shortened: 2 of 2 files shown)


spec.yaml
```diff
@@ -97,7 +97,7 @@ paths:
           },
           "workers": {
             "count_per_cluster": {
-              "max": 20,
+              "max": -1,
               "default": 3
             },
```
xh3b4sd: So the suggestion in the PM was to define the count as `null` for cloud providers. There are multiple values and the `default` is actually useful still IMO. Making a numeric value `null` does not sound quite right to me. So how do we want to express "unlimited" resources? I made it here `-1` as a suggestion. WDYT?

teemow: @xh3b4sd let's maybe go with 500 then. We haven't had clusters that big and probably nobody should create clusters that big. We can adjust this again once we have more information which size of clusters is problematic for our management.

Using `-1` is weird imo.

xh3b4sd: The immediate urge for the PM got already removed by setting it to `100`. It is just that these static numbers are weird and wrong. We should actually have a way to express no limit or indicate the number is unknown. In case you pay for the trusted advisor service you can get some limits I guess and we could add such a support in case the service is available, but it would be nice from the UX point of view to have a mechanism to express that a number is unknown anyway.

tuommaki: `null` was specifically to indicate that this number is unknown. I think within computing it's pretty widely understood to mean undefined or unknown, which unlimited also is, because resources always have a limit, it's just that it's not known what breaks first when there's high enough number of workers (most probably master runs out of resources at that point, but could be also IP address space or something else).

If we implement check with TrustedAdvisor, we should probably then also account for other limits (such as mentioned subnet size and master instance type).

xh3b4sd: Eh but I would like to have some type consistency somehow and `null` implies that you have to deal with that language specifically and everything is just more complicated. In golang we would then need to define an integer pointer or something like that in order to get `nil` and definitely know about `null` being set. Otherwise the result is `0` anyway after unmarshalling and we did not gain anything.

xh3b4sd: See https://play.golang.org/.

tuommaki: Sure, dealing with pointers is slightly cumbersome in this case, but in external interfaces, especially APIs, the clarity and semantic consistency has a lot bigger priority (in my point of view) than code that has to deal with it. This is language specific (and actually library specific, see e.g. how protobuf generates code or how aws-sdk is implemented) implementation detail and shouldn't affect API contract.

spec.yaml
```diff
@@ -97,7 +97,7 @@ paths:
           },
           "workers": {
             "count_per_cluster": {
-              "max": 20,
+              "max": null,
               "default": 3
             },
```
xh3b4sd: I made this `null` now.

marians:
There is no need for all these periods on a description phrase like "General information".

This is fine as it is:

image

spec.yaml
marians: Thanks!

xh3b4sd:
It just broke my heart to see the mess and I tried to bring on some consistency. 😢

marians:
If it makes you happy, there is a rule for it. On signs, in UI labels and in similar places, phrases standing alone don't need a full stop at the end. So we are consistent with the outside world.

No full stops here:

image

image

image

image

image

xh3b4sd:
good_grumpy_cat

marians:
Is this change implemented? We normally merge api-spec PRs after the changes they describe are rolled out.

xh3b4sd:
Oh wow. No it is not. Should I revert it and keep the PR open? I think it will still take some time until the actual implementation lands.

marians:
Yes and add the `waiting-for-implementation` label. Thanks! 🌷

PR #115, opened 2018-10-15
Add /v4/info/app-catalog/

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/4331

This is the spec for the listing of available apps that our apps can eventually use to show users a friendly browsable overview of available apps.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -658,3 +658,59 @@ definitions:
       auth_token:
         type: string
         description: The newly created API token
+
+  V4AppCatalogResponse:
+    type: array
+    items:
+      $ref: '#/definitions/V4AppCatalogItem'
+
+  V4AppCatalogItem:
+    type: object
+    properties:
+      name:
+        type: string
+        description: Unique identifier for this app
+      title:
+        type: string
+        description: User friendly name of this app
+      description:
+        type: string
+        description: A description of what this app this
+      logo_url:
+        type: string
+        description: A url pointing to an image that can be used as an icon or logo for this app
+      screenshots:
+        type: array
+        items:
+          type: object
+          properties:
+            url:
+              type: string
+              description: A url pointing to a screenshot of the app
+            caption:
+              type: string
+              description: A caption describing the screenshot
+      versions:
+        type: array
+        items:
+          type: object
+          properties:
+            version:
+              type: string
+              description: Version number
+            changelog:
+              description: |
+                Structured list of changes in this release, in comparison to the
+                previous version, with respect to the contained components.
+              type: array
+              items:
+                type: object
+                properties:
+                  component:
+                    type: string
+                    description: |
+                      If the changed item was a component, this attribute is the
+                      name of the component.
+                  description:
+                    type: string
+                    description: Human-friendly description of the change
--- a/spec.yaml
+++ b/spec.yaml
@@ -186,7 +186,31 @@ paths:
             $ref: "./definitions.yaml#/definitions/V4AppCatalogResponse"
           examples:
             application/json:
-              []
+              [
+                {
+                  "name": "prometheus",
+                  "title": "Prometheus Monitoring Stack",
+                  "description": "",
+                  "logo_url": "",
+                  "screenshots": [
+                    {
+                      "url": "",
+                      "caption": ""
+                    },
+                  ],
+                  "versions": [
+                    {
+                      "version": "",
+                      "changelog": [
+                        {
+                        "component": "",
+                        "description": ""
+                        }
+                      ],
+                    }
+                  ]
+                }
+              ]
         "401":
           $ref: "./responses.yaml#/responses/V4Generic401Response"
         default:
--- a/spec.yaml
+++ b/spec.yaml
@@ -203,8 +203,8 @@ paths:
                       "version": "",
                       "changelog": [
                         {
-                        "component": "",
-                        "description": ""
+                          "component": "",
+                          "description": ""
                         }
                       ],
                     }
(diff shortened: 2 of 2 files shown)


oponder:
I actually "squashed" together some parts of what would be individual entries per version in the `app.apps` spec we've been building here: https://docs.google.com/document/d/1D2lQveGYbt2qd8nVJ8tW_0kCs6V0-62NrVxZhcyXPrY/edit?ts=5bc062e7

This is motivated more by the kind of listing or page i want to show in Happa, then the underlying data structure we are speccing in the above doc.

Might need a bit more thought.... Perhaps its best to just return the `app.apps` almost as is, and let the clients/front end figure out how to display it nicely as a single coherent page.

xh3b4sd:
I am mostly interested in the response body structure here and from what I know right now I cannot tell if it is "right" or "wrong". I checked the google doc spec but I am not sure where the structure is coming from, why it is as it is and if something is maybe missing or not. We can maybe talk tomorrow in SIG UX sync just FMI. No blocker here though.

definitions.yaml
```diff
@@ -658,3 +658,59 @@ definitions:
...
+          properties:
+            version:
+              type: string
+              description: Version number
+            changelog:
+              description: |
+                Structured list of changes in this release, in comparison to the
+                previous version, with respect to the contained components.
+              type: array
+              items:
```
pipo02mix: I was thinking why not use a link to the release changelog page of the app project?

spec.yaml
```diff
@@ -165,6 +165,59 @@ paths:
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v4/info/app-catalog/:
```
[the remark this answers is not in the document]
puja108: although, no, sorry, taking that back, the catalogs are in the backend, this might be only based on the app-release CRs which are there no matter which catalog their from, which reminds me, we might need a way to track back which catalog an app belongs to, right?

definitions.yaml
```diff
@@ -658,3 +658,59 @@ definitions:
...
+        items:
+          type: object
+          properties:
+            url:
+              type: string
+              description: A url pointing to a screenshot of the app
+            caption:
+              type: string
+              description: A caption describing the screenshot
+      versions:
```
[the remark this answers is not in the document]
puja108: from UX it might not be as nice to have a lot of releases shown, but I would hope we delete old releases quickly

definitions.yaml
```diff
@@ -658,3 +658,59 @@ definitions:
...
+    type: array
+    items:
+      $ref: '#/definitions/V4AppCatalogItem'
+
+  V4AppCatalogItem:
+    type: object
+    properties:
+      name:
+        type: string
+        description: Unique identifier for this app
```
marians: As this is supposed to be a unique identifier, I would go for `id` instead of `name`. Then `title` should be renamed to `name`. That would align it with the cluster resource.

definitions.yaml
```diff
@@ -658,3 +658,59 @@ definitions:
...
+      title:
+        type: string
+        description: User friendly name of this app
+      description:
+        type: string
+        description: A description of what this app this
```
marians: `what this app this`?

definitions.yaml
```diff
@@ -658,3 +658,59 @@ definitions:
...
         type: string
         description: The newly created API token
+
+  V4AppCatalogResponse:
+    type: array
+    items:
+      $ref: '#/definitions/V4AppCatalogItem'
+
+  V4AppCatalogItem:
+    type: object
```
marians: There should be a `required` key telling which attributes are mandatory

spec.yaml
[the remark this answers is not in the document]
marians: How about just `/v4/info/apps/` instead of `/v4/info/app-catalog/`?

definitions.yaml
```diff
@@ -658,3 +658,59 @@ definitions:
...
+        description: Unique identifier for this app
+      title:
+        type: string
+        description: User friendly name of this app
+      description:
+        type: string
+        description: A description of what this app this
+      logo_url:
+        type: string
+        description: A url pointing to an image that can be used as an icon or logo for this app
```
marians: nitpick: `URL`

marians:
I'd need to understand how much content we expect under `versions`. If that will be an ever-growing list of versions, than I'd vote for excluding that from `V4AppCatalogResponse` and instead adding an app details endpoint which would return everything we have for one specific app.

Regarding the `app-catalog` name in the endpoint: I'd like to keep it consistent with our other path names. For a collection of things we normally just use the plural term: `clusters`, `users`, `organizations`, `key-pairs`.

definitions.yaml
puja108: I do want to avoid an artificial ID on purpose if possible, we need to figure this still out though. In k8s this will ne a unique name as the name of a resource can only be unique (at least within a namespace), so maybe there's no need for generating additional unique IDs. I think it would be more beneficial to think about and comment on these things in the overarching design doc here: https://docs.google.com/document/d/1D2lQveGYbt2qd8nVJ8tW_0kCs6V0-62NrVxZhcyXPrY/edit?ts=5bc3ab5b#

puja108: we do have indeed some issue/comment thread about how to do unique naming here that is simple and intuitive also for the time when a user might see it directly. also about how to keep versions apart (separate CRs vs list of versions inside CR)

marians: An ID doesn't have to be artificial. Like in the case of an organization.

oponder:
Closing this as it is outdated. Replacing with a new PR that reflects the latest direction for the App Katalog.

PR #117, opened 2018-10-24
add availability zones

teemow:
clusters can have multiple availability zones now. to get a better
understanding on how this affects subnet ranges and cluster sizes the
info endpoint provides a bit more information about the configured ip
ranges in the installation. clients can then calculate the number of
nodes for the users.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -104,7 +104,7 @@ definitions:
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           to use in the new cluster
       availability_zones:
-        description: Number of availability zones a cluster should be spread across.
+        description: Number of availability zones a cluster should be spread across. The default is provided via the [info](#operation/getInfo) endpoint.
         type: integer
       workers:
         type: array
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -66,7 +66,7 @@ definitions:
                   type: object
                   properties:
                     cidr_blocks:
-                      description: List of CIDRs that will be used by services in Kubernetes
+                      description: List of CIDR blocks that will be used by services in Kubernetes
                       type: array
                       items:
                         type: string
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -39,58 +39,6 @@ definitions:
               default:
                 description: Default number of availability zones for a cluster.
                 type: number
-      network:
-        description: Information related to networking in the clusters
-        type: object
-        properties:
-          ranges:
-            description: IP ranges that can be used by clusters
-            type: array
-            items:
-              type: object
-              properties:
-                name:
-                  description: Name of the IP range (e. g.  "production" or "default")
-                  type: string
-                pods:
-                  description: IP ranges for pods
-                  type: object
-                  properties:
-                    cidr_blocks:
-                      description: List of CIDRs that will be used by pods
-                      type: array
-                      items:
-                        type: string
-                services:
-                  description: IP ranges for services in Kubernetes
-                  type: object
-                  properties:
-                    cidr_blocks:
-                      description: List of CIDR blocks that will be used by services in Kubernetes
-                      type: array
-                      items:
-                        type: string
-                docker:
-                  description: IP ranges for docker
-                  type: object
-                  properties:
-                    cidr_blocks:
-                      description: List of CIDRs that will be used by docker on the nodes
-                      type: array
-                      items:
-                        type: string
-                clusters:
-                  description: IP ranges for clusters.
-                  type: object
-                  properties:
-                    cidr_blocks:
-                      description: List of CIDRs that are available to create clusters.
-                      type: array
-                      items:
-                        type: string
-                    cidr_mask:
-                      description: Mask that is used to cut slices of the cidr_blocks for an individual cluster.
-                      type: number
       workers:
         description: Information related to worker nodes
         type: object
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -34,7 +34,7 @@ definitions:
             description: Number of availability zones which a cluster can be spread across.
             properties:
               max:
-                description: Maximum number of availability zones in this region.
+                description: Number of availability zones in the region of this installation.
                 type: number
               default:
                 description: Default number of availability zones for a cluster.
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -50,7 +50,7 @@ definitions:
               type: object
               properties:
                 name:
-                  description: Name of the IP range (eg production or default)
+                  description: Name of the IP range (e. g.  "production" or "default")
                   type: string
                 pods:
                   description: IP ranges for pods
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -31,7 +31,7 @@ definitions:
             type: string
           availability_zones:
             type: object
-            description: Number of availability zones that the cluster can be spread across.
+            description: Number of availability zones which a cluster can be spread across.
             properties:
               max:
                 description: Maximum number of availability zones in this region.
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -29,6 +29,68 @@ definitions:
           datacenter:
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
+          availability_zones:
+            type: object
+            description: Number of availability zones that the cluster can be spread across.
+            properties:
+              max:
+                description: Maximum number of availability zones in this region.
+                type: number
+              default:
+                description: Default number of availability zones for a cluster.
+                type: number
+      network:
+        description: Information related to networking in the clusters
+        type: object
+        properties:
+          ranges:
+            description: IP ranges that can be used by clusters
+            type: array
+            items:
+              type: object
+              properties:
+                name:
+                  description: Name of the IP range (eg production or default)
+                  type: string
+                pods:
+                  description: IP ranges for pods
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that will be used by pods
+                      type: array
+                      items:
+                        type: string
+                services:
+                  description: IP ranges for services in Kubernetes
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that will be used by services in Kubernetes
+                      type: array
+                      items:
+                        type: string
+                docker:
+                  description: IP ranges for docker
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that will be used by docker on the nodes
+                      type: array
+                      items:
+                        type: string
+                clusters:
+                  description: IP ranges for clusters.
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that are available to create clusters.
+                      type: array
+                      items:
+                        type: string
+                    cidr_mask:
+                      description: Mask that is used to cut slices of the cidr_blocks for an individual cluster.
+                      type: number
       workers:
         description: Information related to worker nodes
         type: object
@@ -86,6 +148,9 @@ definitions:
         description: |
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           to use in the new cluster
+      availability_zones:
+        description: Number of availability zones a cluster should be spread across.
+        type: number
       workers:
         type: array
         items:
@@ -140,6 +205,9 @@ definitions:
         description: |
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           currently running this cluster.
+      availability_zones:
+        description: Number of availability zones a cluster is spread across.
+        type: number
       workers:
         type: array
         items:
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -42,10 +42,10 @@ definitions:
             properties:
               max:
                 description: Number of availability zones in the region of this installation.
-                type: number
+                type: integer
               default:
                 description: Default number of availability zones for a cluster.
-                type: number
+                type: integer
       workers:
         description: Information related to worker nodes
         type: object
@@ -105,7 +105,7 @@ definitions:
           to use in the new cluster
       availability_zones:
         description: Number of availability zones a cluster should be spread across.
-        type: number
+        type: integer
       workers:
         type: array
         items:
@@ -162,7 +162,7 @@ definitions:
           currently running this cluster.
       availability_zones:
         description: Number of availability zones a cluster is spread across.
-        type: number
+        type: integer
       workers:
         type: array
         items:
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -15,10 +15,14 @@ definitions:
   # Info response
   V4InfoResponse:
     type: object
+    required:
+      - general
     properties:
       general:
         description: General information
         type: object
+        required:
+          - availability_zones
         properties:
           installation_name:
             description: Unique name of the installation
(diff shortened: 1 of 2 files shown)


definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
+          ranges:
+            description: IP ranges that can be used by clusters
+            type: array
+            items:
+              type: object
+              properties:
+                name:
+                  description: Name of the IP range (eg production or default)
+                  type: string
+                pods:
```
MarcelMue: Any reason for this order? `clusters`, `docker`, `pods`, `services` would be alphabetical. Am I missing something?

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
+          availability_zones:
+            type: object
+            description: Number of availability zones that the cluster can be spread across.
+            properties:
+              max:
+                description: Maximum number of availability zones in this region.
+                type: number
+              default:
```
MarcelMue: The default should always be `1`, so this makes sense.

definitions.yaml
teemow: I don't know. Any reason to sort it alphabetical?

From a scope point of view its probably:

* clusters
* services
* pods
* docker

Cluster API also specifies services and then pods: https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/cluster_types.go#L58

In fact. I don't care! :smile:

MarcelMue: Well that would make sense to me as well, currently it is clusters last though - which is why I wondered this :D

spec.yaml
```diff
@@ -143,7 +187,29 @@ paths:
                 "general": {
                   "installation_name": "shire",
                   "provider": "aws",
-                  "datacenter": "eu-central-1"
+                  "datacenter": "eu-central-1",
+                  "availability_zones": {
+                    "max": 3,
+                    "default": 1
+                  }
```
xh3b4sd: Can we add the actual list of values as well? So far I thought we use the info endpoint internally as well which would make the work with it easier. Otherwise there has to be a separate mechanism to get them anyway.

spec.yaml
```diff
@@ -819,6 +911,7 @@ paths:
         ### Limitations
 
         - As of now, existing worker nodes cannot be modified.
+        - The number of availability zones can not be changed afterwards.
```
xh3b4sd: The line above has `cannot`. Here we have `can not` now. Would be cool to align.

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
+                        type: string
+                docker:
+                  description: IP ranges for docker
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that will be used by docker on the nodes
+                      type: array
+                      items:
+                        type: string
```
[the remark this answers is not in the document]
teemow: yes you are right and i don't like it here as well. imo we should get rid of it completely but this might not be easy. 

my thinking here is: 
this range is configured on each node. the bridge is there and this ip range can't be routed elsewhere. so this is of interest for the user. we would need to get rid of the docker bridge then. its not used anyway imo (at least on the nodes).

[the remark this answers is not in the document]
puja108: we might move to CRI and containerd at some time anyway, then we would not even need to have the daemon there 😉

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
           datacenter:
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
+          availability_zones:
+            type: object
+            description: Number of availability zones that the cluster can be spread across.
```
```suggestion
            description: Number of availability zones which a tenant cluster can be spread across.
```

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
+            properties:
+              max:
+                description: Maximum number of availability zones in this region.
```
```suggestion
                description: Number of availability zones in the region of this installation.
```

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
+        type: object
+        properties:
+          ranges:
+            description: IP ranges that can be used by clusters
+            type: array
+            items:
+              type: object
+              properties:
+                name:
+                  description: Name of the IP range (eg production or default)
```
```suggestion
                  description: Name of the IP range (e. g.  "production" or "default")
```

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
+                clusters:
+                  description: IP ranges for clusters.
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that are available to create clusters.
+                      type: array
+                      items:
+                        type: string
+                    cidr_mask:
```
teemow: btw i left out the public/private mask that we currenlty also configure in each installation because this needs to be calculated based on the amount of availability zones.

spec.yaml
teemow: what do you mean? the names of the availability zones? why do we need them?

definitions.yaml
```diff
@@ -29,6 +29,68 @@ definitions:
...
+                      description: List of CIDRs that will be used by pods
+                      type: array
+                      items:
+                        type: string
+                services:
+                  description: IP ranges for services in Kubernetes
+                  type: object
+                  properties:
+                    cidr_blocks:
+                      description: List of CIDRs that will be used by services in Kubernetes
```
```suggestion
                      description: List of CIDR blocks that will be used by services in Kubernetes
```

definitions.yaml
teemow: :heart:

marians:
A few general notes:

If we think that we might change the API soon so that instead of a number of availability zones users should define specifically which ones a cluster should use, then we should go for a more extensible API design that allows to prevent breaking changes. In cluster creation:

```
"availability_zones": {
  "number": 3
}
```

could be later extended to allow, as an alternative, the one below, while still supporting the above:

```
"availability_zones": {
  "selection": ["eu-west-1a", "eu-west-1c"]
}
```

A style detail: Alle the little `description` values do not need a `.` in the end, as long as they are only one sentence long.

spec.yaml
xh3b4sd: We need them latest in the `aws-operator`. So where do we get them from then? My thinking was that we simply use the information from the info endpoint internally as well and it would also be more informational for the customer when using the endpoint.

teemow:
> If we think that we might change the API soon so that instead of a number of availability zones users should define specifically which ones a cluster should use, then we should go for a more extensible API design that allows to prevent breaking changes

No this won't be the case.

spec.yaml
xh3b4sd: I read the randomization spec and I like it. We can do this. This is separate from having the real AZ names in the info endpoint though. Why should we not expose them here? Having information in the info endpoint is pretty good IMO.

teemow: I see no reason why we should expose them.

xh3b4sd: I gave you a reason already. We can use the information provided by the info endpoint internally because we need these information. We can use the info endpoint on our own as interface. I would like to have the AZ names in here for that purpose. The alternative is to have some other mechanism which is then some other interface and we could just make it easier and dog food that way.

teemow: > We can use the information provided by the info endpoint internally because we need these information

Internal reason don't justify exposing this externally through the api. and i don't see why something like cluster-service should suddenly connect via the api. let keep things minimal. It is not needed in the interface and thats why it won't be exposed. 

the other problem i have with this is that we expose information that would lead to false assumptions. if you specify one availability zone it won't end up in the first AZ. of course we can describe this. maybe we should provide more info about this.

xh3b4sd: Disagree and commit.

definitions.yaml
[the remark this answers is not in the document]
r7vme: +1 to remove docker ranges now.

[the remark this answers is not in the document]
teemow: we might remove this entirely. still discussing here: https://github.com/giantswarm/giantswarm/pull/2202#issuecomment-433038314

any opinion about omiting this and waiting for https://github.com/giantswarm/giantswarm/issues/4490 instead?

spec.yaml
teemow: If those things are bugging you. There is now a nice way how you can suggest changes. Marian did this already in this PR: https://github.com/giantswarm/api-spec/pull/117#discussion_r228089368

xh3b4sd:
```suggestion
        - The number of availability zones cannot be modified afterwards.
```

teemow: :heart:

spec.yaml
```diff
@@ -601,6 +615,28 @@ paths:
...
 
+        The number of `availability_zones` affects the total number of nodes
+        that can be created in the cluster. The number of availability zones
+        splits the ip range that can be used for the cluster in multiple smaller
+        ip ranges. The [getInfo](#operation/getInfo) endpoint provides more
+        details about the cluster ip range.
+
+        IP range example:
+
+        If a cluster gets a `/22` (1022 hosts). And the cluster should be
```
tuommaki: Maybe remove period before and: `If a cluster gets a `/22` (1022 hosts) and the cluster...`

spec.yaml
```diff
@@ -601,6 +615,28 @@ paths:
...
+        spawned across 3 availability zones. The range will then be split up
+        into four `/24` (254 hosts) that can be given to four different
+        availability zones. One range will be stay unused because ranges can
```
tuommaki: Maybe reword a bit: `One range will be unused because network addresses must be powers of two. See [CIDR addressing](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing).`

spec.yaml
```diff
@@ -601,6 +615,28 @@ paths:
...
+        only be split binary. Each of the `/24` will then be split into two
+        `/25` (126 hosts) for public and private subnets. The private subnet is
+        used for nodes and internal loadbalancer (only if you create them
+        within Kubernetes). The public subnet will be used by the public
+        loadbalancers. Your cluster comes with two public loadbalancers already.
```
tuommaki: `Tentant clusters come with two public loadbalancers by default.`

spec.yaml
```diff
@@ -601,6 +615,28 @@ paths:
...
+        One for the Kubernetes API and one for Ingress.
+
+        __Note:__ AWS ELBs can take up to 8 IP addresses due to the way how
+        they scale.
```
tuommaki: `they scale. In addition to this, every AWS subnet has four first addresses (.1-.4) reserved for internal use.`

spec.yaml
```diff
@@ -93,7 +93,11 @@ paths:
           "general": {
             "installation_name": "shire",
             "provider": "aws",
-            "datacenter": "eu-central-1"
+            "datacenter": "eu-central-1",
```
tuommaki: `datacenter` field seems to have been there already earlier, but it somehow struck me now that this is wrong. Each AZ is a datacenter and `datacenter` here should most probably be a `region`. Probably worth handling separately, but just mentioning here as I happened to notice this.

teemow: :+1: but lets do this separately.

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
         omitted, default configuration values will be applied. For example, if
         no `release_version` is specified, the most recent version is used.
 
+        The number of `availability_zones` affects the total number of nodes
+        that can be created in the cluster. The number of availability zones
+        splits the ip range that can be used for the cluster in multiple smaller
```
marians:
```suggestion
        splits the IP range that can be used for the cluster in multiple smaller
```

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
...
+        ip ranges. The [getInfo](#operation/getInfo) endpoint provides more
```
marians:
```suggestion
        IP ranges. The [getInfo](#operation/getInfo) endpoint provides more
```

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
...
+        details about the cluster ip range.
```
marians:
```suggestion
        details about the cluster IP range.
```

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
...
+
+        IP range example:
+
+        If a cluster gets a `/22` (1022 hosts) and the cluster should be
+        spawned across 3 availability zones. The range will then be split up
+        into four `/24` (254 hosts) that can be given to four different
```
marians:
```suggestion
        into four `/24` (254 hosts) that can be assigned to four different
```

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
...
+        availability zones. One range will be stay unused because network
```
marians:
```suggestion
        availability zones. One range will stay unused because network
```

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
...
 
+        The number of `availability_zones` affects the total number of nodes
+        that can be created in the cluster. The number of availability zones
+        splits the ip range that can be used for the cluster in multiple smaller
+        ip ranges. The [getInfo](#operation/getInfo) endpoint provides more
+        details about the cluster ip range.
+
+        IP range example:
+
+        If a cluster gets a `/22` (1022 hosts) and the cluster should be
```
marians:
```suggestion
        If a cluster gets a `/22` range (1022 hosts) and the cluster should be
```

spec.yaml
```diff
@@ -605,6 +619,30 @@ paths:
...
+        spawned across 3 availability zones. The range will then be split up
```
marians:
```suggestion
        spawned across 3 availability zones, the range will then be split up
```

definitions.yaml
```diff
@@ -86,6 +103,9 @@ definitions:
         description: |
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           to use in the new cluster
+      availability_zones:
+        description: Number of availability zones a cluster should be spread across.
```
marians: Can we inform here that a default will be applied (as given in info endpoint -> `general. availability_zones.default`)?

definitions.yaml
```diff
@@ -86,6 +103,9 @@ definitions:
...
+        type: number
```
MarcelMue:
```suggestion
        type: integer
```
This currently generates to a `float64` in `gsclientgen` but should instead be an `int`

definitions.yaml
```diff
@@ -140,6 +160,9 @@ definitions:
         description: |
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           currently running this cluster.
+      availability_zones:
+        description: Number of availability zones a cluster is spread across.
+        type: number
```
MarcelMue:
```suggestion
        type: integer
```

definitions.yaml
```diff
@@ -29,6 +33,19 @@ definitions:
...
+            required:
+              - max
+              - default
+            properties:
+              max:
+                description: Number of availability zones in the region of this installation.
+                type: number
+              default:
+                description: Default number of availability zones for a cluster.
+                type: number
```
MarcelMue:
```suggestion
                type: integer
```

definitions.yaml
```diff
@@ -29,6 +33,19 @@ definitions:
...
+          availability_zones:
+            type: object
+            description: Number of availability zones which a cluster can be spread across.
+            required:
+              - max
+              - default
+            properties:
+              max:
+                description: Number of availability zones in the region of this installation.
+                type: number
```
MarcelMue:
```suggestion
                type: integer
```

PR #118, opened 2018-11-07
Adds note to AuthorizationHeaderToken about SSO

JosephSalisbury:
This bit me earlier, it should be documented.

screen shot 2018-11-07 at 14 14 22

becoming:

screen shot 2018-11-08 at 11 03 57

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -65,6 +65,10 @@ securityDefinitions:
       header with a value of the format `giantswarm <token>`. Auth tokens can be
       obtained using the [createAuthToken](#operation/createAuthToken)
       operation.
+
+      For Giant Swarm staff who are authenticating using SSO,
+      pass the auth token via the `Authorization` header with a value of the
+      format `Bearer <token>`.
     type: apiKey
     name: Authorization
     in: header


JosephSalisbury:
Let's see what the rest of @giantswarm/sig-ux says

oponder:
I vote for not documenting this in the api-spec until users can actually make use of it.

Or otherwise noting that this is just for 'giant swarm admins', to avoid that people go wild goose chase trying to figure out how to configure SSO for themselves, or get the wrong expectation.

JosephSalisbury:
Being explicit that it's for Giant Swarm admins would work for me - I went to the api docs to work out how to do this, so I'd say I'm a user that can actually make use of it :D

JosephSalisbury:
```
For Giant Swarm staff who are authenticating using SSO, pass the auth token via the `Authorization` header with a value of the format `Bearer <token>`.
```

@marians would that work for you? I think it makes it explicit that this is only for Giant Swarm staff.

marians:
That would work for me

JosephSalisbury:
@marians RFR

PR #119, opened 2018-11-22
add availability zone to cluster details and listing

teemow:
respond with the actual list of availability zones of a cluster.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -161,8 +161,11 @@ definitions:
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           currently running this cluster.
       availability_zones:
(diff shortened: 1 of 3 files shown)


definitions.yaml
```diff
@@ -523,6 +526,12 @@ definitions:
       owner:
         type: string
         description: Name of the organization owning the cluster
+      availability_zones:
```
MarcelMue: Unsure if this is necessary in the list of clusters. Kind of a UX decision but I don't see a list of possibly 3+ AZ names as useful here.

pipo02mix: Maybe to add a column with the number of AZs

PR #121, opened 2018-11-30
Support autoscaling of tenant clusters

xh3b4sd:

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -106,6 +106,20 @@ definitions:
       availability_zones:
         description: Number of availability zones a cluster should be spread across. The default is provided via the [info](#operation/getInfo) endpoint.
         type: integer
+      scaling:
+        type: object
+        description: |
+          Attributes specific to cluster node scaling. To have full control of
+          cluster scaling, min and max can be set to the same value.
+        properties:
+          min:
+            type: integer
+            description: |
+              The minimum number of cluster nodes
+          max:
+            type: integer
+            description: |
+              The maximum number of cluster nodes
       workers:
         type: array
         items:
@@ -166,6 +180,25 @@ definitions:
         items:
           description: Name of the availability zone
           type: string
+      scaling:
+        type: object
+        description: |
+          Attributes specific to cluster node scaling. To have full control of
+          cluster scaling, min and max can be set to the same value.
+        properties:
+          current:
+            type: integer
+            description: |
+              The current number of cluster nodes. Most insightful when min and
+              max define a range in which the autoscaler operates autonomously.
+          min:
+            type: integer
+            description: |
+              The minimum number of cluster nodes
+          max:
+            type: integer
+            description: |
+              The maximum number of cluster nodes
       workers:
         type: array
         items:
@@ -221,20 +254,6 @@ definitions:
           size_gb:
             type: number
             description: RAM size in GB. Can be an integer or float.
-      scaling:
-        type: object
-        description: |
-          Attributes specific to cluster node scaling. To have full control of
-          cluster scaling, min and max can be set to the same value.
-        properties:
-          min:
-            type: integer
-            description: |
-              The minimum number of cluster nodes in this node pool
-          max:
-            type: integer
-            description: |
-              The maximum number of cluster nodes in this node pool
       storage:
         type: object
         properties:
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -110,7 +110,7 @@ definitions:
         type: object
         description: |
           Attributes specific to cluster node scaling. To have full control of
-          cluster scaling, min and max can be set to the same value.
+          the cluster size, min and max can be set to the same value.
         properties:
           min:
             type: integer
@@ -142,7 +142,7 @@ definitions:
         type: object
         description: |
           Attributes specific to cluster node scaling. To have full control of
-          cluster scaling, min and max can be set to the same value.
+          the cluster size, min and max can be set to the same value.
         properties:
           min:
             type: integer
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -122,6 +122,9 @@ definitions:
               The maximum number of cluster nodes
       workers:
         type: array
+        description: |
+          Worker node definition. If present, the first item of the array is
+          expected to contain the specification for all worker nodes.
         items:
           $ref: '#/definitions/V4NodeDefinition'
 
@@ -152,6 +155,11 @@ definitions:
             type: integer
             description: |
               The maximum number of cluster nodes
+      workers:
+        type: array
+        description: Worker node array (deprecated)
+        items:
+          $ref: '#/definitions/V4NodeDefinition'
 
   # Details on existing cluster
   V4ClusterDetailsResponse:
@@ -190,6 +198,7 @@ definitions:
           description: Name of the availability zone
           type: string
       workers:
+        description: Information about worker nodes in the cluster (deprecated)
         type: array
         items:
           $ref: '#/definitions/V4NodeDefinition'
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -152,11 +152,6 @@ definitions:
             type: integer
             description: |
               The maximum number of cluster nodes
-      workers:
-        type: array
-        description: Worker node array
-        items:
-          $ref: '#/definitions/V4NodeDefinition'
 
   # Details on existing cluster
   V4ClusterDetailsResponse:
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -221,6 +221,20 @@ definitions:
           size_gb:
             type: number
             description: RAM size in GB. Can be an integer or float.
+      scaling:
+        type: object
+        description: |
+          Attributes specific to cluster node scaling. To have full control of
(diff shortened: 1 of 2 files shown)


definitions.yaml
```diff
@@ -221,6 +221,20 @@ definitions:
...
+          cluster scaling, min and max can be set to the same value.
+        properties:
+          min:
+            type: integer
+            description: |
+              The minimum number of cluster nodes in this node pool
+          max:
+            type: integer
+            description: |
+              The maximum number of cluster nodes in this node pool
```
xh3b4sd: I thought a bit about the API and what to do here for cluster creation, especially having node pools in mind. The API is already quite aligned with that I think. We have the `workers` list which contain all the node settings. Later we might just rename this to `nodes` or whatever. So now I just added the scaling section. Naming is up for discussion. Is this good enough for cluster creation?

tuommaki: So do I understand this correctly that now `scaling: {min/max}` parameter is property of a single worker node and we use it like instance type etc. right now? This is definitely unintuitive to me (I'd rather have it in one step higher level) but as you write, this aligns with what we currently have. I don't care so much about this right now though, when we are moving away from these with node pools.

teemow: Why should the `scaling` be defined per node? This doesn't make sense. I know the worker is already wrong in general as we only use the first worker configuration to create a pool of nodes.

Since we know already that this is gonna change (and fixed) with node pools and cluster api I am fine with keeping the worker list as is. But the `scaling` should go into the `V4ClusterDetailsResponse` right next to eg the availability zones.

marians:
The nice thing about Tim's proposal is that it's a non-breaking change. However, I agree that with having a `scaling` attribute on every node we will end up with a representation that doesn't fit really. So I would rather vote for a breaking change (which has to be communicated well to customers, and we might want to talk about ways to ease the transition).

Once we implement node pools, will scaling become an attribute of a node pool, so that each pool can be scaled independently? I hope yes.

In that case, the rough form could be like

```json
{
  ...
  "node_pools": [
    {
      "aws": {"instance_type": "m5.2xlarge"},
      "scaling": {"min": 3, "max": 10}
    }
]
}

So scaling is not a cluster attribute, but one of the node pool. We would only support one entry in the `node_pools` list in the beginning, adding support for more later.

**Comment by teemow** [comment]
@marians yes autoscaling will be defined per node pool. see: https://github.com/giantswarm/giantswarm/pull/2222/files#diff-9ba784d19773fc33a28f6782bfefe102R8

I am not certain how the api will look like yet. I'd like to align this more to the k8s structures. See https://github.com/giantswarm/apiextensions/pull/173#issuecomment-442923469

Introducing a `node_pools` structure is imo out of scope for the autoscaling story.

File: definitions.yaml (hunk 1 above)
Comment (xh3b4sd) [reply definitions.yaml]: My thinking was that we use the workers list here as node pool list in the future and that you want to define the scaling settings per pool.

**Comment by pipo02mix** [comment]
But Timo, what is wrong, only the naming or the structure?

It is only the api spec (api gateway) and it will be changed when we moved to `ClusterAPI`. The interface will totally difference (k8s resources). I mean, we can convert in the API gateway this naming/structure to k8s Cluster API terms and structure

File: definitions.yaml [hunk 2]
```diff
@@ -166,6 +180,25 @@ definitions:
         items:
           description: Name of the availability zone
           type: string
+      scaling:
+        type: object
+        description: |
+          Attributes specific to cluster node scaling. To have full control of
+          cluster scaling, min and max can be set to the same value.
+        properties:
+          current:
```
xh3b4sd: I did the same for the `V4ClusterDetailsResponse` now. Here I also added a `current` field. WDYT?

teemow: @xh3b4sd the plan was to start separating desired and actual state better in our api. 

Please see: https://github.com/giantswarm/giantswarm/pull/2206/files#diff-c512236ee39b9619d47c35ed63f371a4R51

xh3b4sd: I was not aware of that. Thank you for pointing this out. The new status endpoint sounds good, but is like a new feature to me. Do we want to sort this in a separate story or is this what we have to do now? Just saying, this increases complexity and makes the story take longer to be addressed.

teemow: it is part of the autoscaling story. we need to provide the current number of nodes and therefore we anyway have to read the status. so the idea is to expose all information we have in the status resource. but happy to have a chat about it. hangout might be easier.

teemow: @xh3b4sd and I decided that the status goes into a separate PR.

spec.yaml
```diff
@@ -675,6 +675,7 @@ paths:
                 "release_version": "1.4.2",
                 "name": "Example cluster with 3 default worker nodes",
                 "availability_zones": 2,
+                "scaling": { "min": 3, "max": 3 },
```
MarcelMue: There was an fmt error here because the spaces inside these brackets I believe.

xh3b4sd: Fixed.

teemow:
adapting the scaling endpoint to change min/max could make sense imo. but how is this endpoint different from the general update endpoint? @marians wdyt?

marians:
> Right now we have a dedicated scaling endpoint

Scaling in practice happens via `PATCH /v4/cluster/{cluster-id}/`. There is no scaling endpoint currently.

marians:
As the spec draft looks right now, I think modification of the scaling parameters should still be done via `PATCH /v4/clusters/{cluster-id}/`. With JSON merge patch, this can be a request as simple as these (examples):

```nohighlight
PATCH /v4/clusters/{cluster-id}/
Content-type: application/merge-patch+json

{
  "scaling": {"min": 10}
}
```

```nohighlight
PATCH /v4/clusters/{cluster-id}/
Content-type: application/merge-patch+json

{
  "scaling": {"min": 10, "max": 10}
}
```

```nohighlight
PATCH /v4/clusters/{cluster-id}/
Content-type: application/merge-patch+json

{
  "scaling": {"max": 10}
}
```

xh3b4sd:
You are right. I am not sure where I got the idea from that we have a dedicated scaling endpoint. Probably wishful thinking. I will add the scaling settings to the cluster update endpoint.

spec.yaml
```diff
@@ -828,6 +829,10 @@ paths:
         cluster to a newer
         [release](https://docs.giantswarm.io/api/#tag/releases).
 
+        - `scaling`: Adjust the cluster node limits to make use of auto scaling
+        or to have full control over the node count. The latter can be
+        achieved by setting min and max to the same values.
+
         - `workers`: By modifying the array of workers, nodes can be added to
```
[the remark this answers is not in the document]
marians: As discussed in the SIG UX sync, I'll work on this. Will start a PR based on this one.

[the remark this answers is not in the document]
xh3b4sd: Much love.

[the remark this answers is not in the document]
teemow: That is a good question.

We need to refactor all places in which we use the length of the array as the number of workers anyway to replace it with the scaling settings. And afaik operators only look at the first worker anyway to determine instance settings for the whole cluster.

So either

1) We leave it as is, as it is kinda broken already anyway and the real change will come with node pools and cluster api.
2) We remove the array and just have instance settings.

@marians please sync with me before you start any additional work here.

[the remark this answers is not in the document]
xh3b4sd: I would like to just ignore the workers list here. It is the easiest for us for now. When we document that and make it clear that way I am fine. As you said, it was not really working nor useful anyway.

teemow:
A few things we need to be aware of:

 * we will have older releases of clusters that do not have the scaling attribute yet
 * we will have providers that don't support autoscaling yet

As a consequence:
 * we need to make the clients aware that sometimes autoscaling isn't supported and min=max or max is ignored
 * we need to make sure that scaling of clusters based on older releases still works (without the autoscaler)
 * we should default to sane values. if not too much effort we should make sure that cluster creation with an older version of gsctl still works. but maybe this isn't fully possible. (eg removing the workers from the cluster detail response is definitely a breaking change).

pipo02mix:
Personally, I would not remove `workers` from the response without increase the version. (Some customers can be using these entities)

> we need to make sure that scaling of clusters based on older releases still works (without the autoscaler)

I thought if we use the same value for `min` and `max` then old scaling should work as before. So for old clusters should work the same way as we have the same value (maybe we can enforce it in validation side)

marians:
In https://github.com/giantswarm/api-spec/pull/121/commits/0901fd26eeca95b9e0ec521c55b3f08accb55ace I modified the info regarding the `workers` array again for

- creating a cluster
- modifying a cluster (scaling)
- getting cluster details

Please give this a bullshit check. Thanks!

xh3b4sd:
Thanks for your additions Marian. Makes sense to me. Just fixed some validation issues.

xh3b4sd:
I noticed that we did not yet define the extensions for the cluster details response. Should we return the `scaling` settings when cluster details are fetched? I thought we had this sorted already but the spec misses this.

pipo02mix:
I think it will be needed to show the scaling setting on the cluster detail page or gsctl cluster info, right?

PR #122, opened 2018-12-04
adds spec for status endpoint

xh3b4sd:

Towards https://github.com/giantswarm/giantswarm/pull/2206.


definitions.yaml
```diff
@@ -237,6 +237,114 @@ definitions:
...
+                  description: Last time the condition was checked and verified
+                lastTransitionTime:
+                  type: string
+                  description: |
+                    Last time the condition transitioned from one state to
+                    another
+                status:
+                  type: string
+                  description: |
+                    Status of the condition. One of True, False, or Unknown.
```
teemow: Maybe also add what status of True, False or Unknown means.

definitions.yaml
```diff
@@ -237,6 +237,114 @@ definitions:
...
+                            Public subnet CIDR allocated for this availability
+                            zone
+      cluster:
+        type: object
+        properties:
+          conditions:
+            type: array
+            items:
+              type: object
+              description: TODO
```
teemow: How about: "List of cluster conditions. Cluster conditions describe the state transitions of a cluster (eg. cluster creation, upgrade and deletion)."

Maybe also giving an example of a condition here. Eg a cluster was created or is currently upgraded.

marians:
Thanks for the info!

marians:
Please understand that this is tough for me now. I guess I already know why you're proposing it the way you do, anyway I have to point out the problems in the context of the API.

image

Why have a `cluster` key and a `aws` key on the top level?

This is the cluster status endpoint. Having it contain a `cluster` key on the top level "stutters", as golint would say.

If we would copy the design we have in the cluster detials endpoint we would take the keys that are now in the `cluster` object and would move them to the top. Then we would also have a provider-specific key like `aws`, `azure`, `kvm`.

Why have an `availabilityZones` key under `aws`?

In the cluster creation request and cluster details response, the `availabiltiy_zones` key is a cluster-level attribute, not provider specific.

And there is the `camelCase` vs. `underscore_case` style difference.

I know this stems from the desire to simply take parts from two CRs mostly unmodified and publish them. We can do that, but it doesn't make for a great API design and it will definitely raise some people's eyebrows.

definitions.yaml
xh3b4sd: Feel free to suggest a concrete edit here that pleases your personal taste. I do not know what may be right or wrong here.

definitions.yaml
xh3b4sd: I added some brief description, but as above the example is missing. Please suggest something.

marians:
Tim and I spoke. We decided to make it very clear and obvious that the ../status/ endpoint really just passes through what the CR's status contains. This includes the camelCase, and `resources`. This way we can also point users to `apiextensions` for further docs.

@teemow We spoke about this topic in SIG UX today with a slightly different outcome (not copying 100% what we have in apiextensions). Now it wasn't exactly clear to me that the result of exposing the CR's status would mean what we have now (especially `aws.availabilityZones`, `cluster.*`).

We can favour agility over stability and consistency here, but we should try to manage user's expectations. I offered Tim to provide a preamble text for this endpoint to help with that.

teemow:
@marians what about godoc? 

https://godoc.org/github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1#AWSConfig

marians:
Very good, thanks! Links added.

spec.yaml
```diff
@@ -1039,6 +1039,109 @@ paths:
...
+      tags:
+        - clusters
+      summary: Get cluster status
+      description: |
+        Returns an object about a cluster's current state and past status transitions.
+
+        This endpoint passes through the status content of the Kubernetes resources representing
+        a cluster in the according custom resource. That is, depending on the provider:
+
+        - [awsconfig.provider.giantswarm.io](https://godoc.org/github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1#AWSConfig)
```
teemow: We might even want to link directly to the Status resource here: 

https://godoc.org/github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1#AWSConfigStatus

marians: I didn't do it now, because I think it's better to point to the CR we mention. The users will find their way there though.

definitions.yaml
```diff
@@ -237,6 +237,154 @@ definitions:
...
+                    Condition type describing what kind of condition this is
+                    about. One of Creating, Created, Updating, Updated, or
+                    Deleting.
+          network:
+            type: object
+            properties:
+              cidr:
+                type: string
+                description: |
+                  The whole network CIDR for the cluster. Divided into multiple
```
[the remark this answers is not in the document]
marians: FYI: In our API docs it's fairly common to only talk about a "cluster" when we in fact mean the tenant cluster, because there is hardly any reason to talk about the control plane in this context.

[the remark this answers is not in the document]
tuommaki: Oh, then this suggestion is not correct. Ignore it :flushed: 

Thanks for notifying about this!

xh3b4sd:
Is this spec fine for you? Please let me know so we can start the implementation.

marians:
I am still not sure whether it is the right way to include the status model into the spec, because of the version-based changes we are expecting there. We could basically only support one version and would let all other versions fail. I'd like to hear @oponder's opinion.

An alternative would be to specify the entire model for that endpoint freely using `additionalProperties: true` (https://swagger.io/docs/specification/data-models/dictionaries/#free-form).

This would mean that clients using the spec could use the generated function to pull the status for a cluster, including authentication, but would not get support for parsing the info returned. This leaves the work to deal with schema changes between releases to the client implementations.

xh3b4sd:
The status structure here is a union of all providers. Certain fields will be empty. I do not see the problem here.

teemow:
@marians if i understand correctly you are worried because the schema of this endpoint is based on the version of the cluster, correct? but we will only use the latest apiextensions and therefore fields will only be empty like @xh3b4sd said. so the only thing we need to take care of in the future is to adapt this schema here everytime we change the status resource in apiextensions.

oponder:
I'm not sure about the versioning situation, if it works as timo says, then I have no problems here.

We have to make sure to make backwards compatible changes when making new versions of the CR in apiextensions, or otherwise increment the GS API version, though chances are it won't come to that, and the GS API will be deprecated by then.

marians:
As discussed in the SIG-UX meeting we'd like to make it clear that the `apiextensions` repo is the source of truth for the schema of the status endpoint. And avoid duplicating the schema definition. So here we will try using only `additionalProperties: true` in the status model definition.

marians:
If we want to, we can add an example via `responses` -> `examples` -> `application/json`. But this is something that has to be maintained, too.

You find plenty of examples for this in `spec.yaml`, e. g. in `getClusters`.

xh3b4sd:
There was a spec. There was an example. We said we do not want to maintain in two places. So I removed both. When we keep the example we should keep the spec. As I said, I do not mind which way to go, but mixing this seems to be super wrong to me.

marians:
Cool, then leave it out. You stated you found something odd. I spelled out what I think the options are.

xh3b4sd:
The `api` PR got merged. The status endpoint is implemented. Can we merge this here then? Is Marian happy? And everybody else?

teemow:
LGTM

marians:
Cool! Here is a real-world example.

```json
{
  "aws": {
    "availabilityZones": null
  },
  "cluster": {
    "conditions": [
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "status": "True",
        "type": "Updated"
      }
    ],
    "network": {
      "cidr": ""
    },
    "nodes": [
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "name": "6iec4-master-000000",
        "version": "2.0.1"
      },
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "name": "6iec4-worker-000002",
        "version": "2.0.1"
      },
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "name": "6iec4-worker-000003",
        "version": "2.0.1"
      },
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "name": "6iec4-worker-000004",
        "version": "2.0.1"
      },
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "name": "6iec4-worker-000009",
        "version": "2.0.1"
      },
      {
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "name": "6iec4-worker-00000e",
        "version": "2.0.1"
      }
    ],
    "resources": [
      {
        "conditions": [
          {
            "lastHeartbeatTime": "0001-01-01T00:00:00Z",
            "lastTransitionTime": "0001-01-01T00:00:00Z",
            "status": "InstancesUpgrading",
            "type": "Stage"
          }
        ],
        "name": "instancev4"
      }
    ],
    "scaling": {
      "desiredCapacity": 0
    },
    "versions": [
      {
        "date": "2018-11-15T13:37:39.863296744Z",
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "semver": "2.0.0"
      },
      {
        "date": "2018-12-06T11:22:15.474857203Z",
        "lastHeartbeatTime": "0001-01-01T00:00:00Z",
        "lastTransitionTime": "0001-01-01T00:00:00Z",
        "semver": "2.0.1"
      }
    ]
  }
}
```

PR #123, opened 2018-12-11
workers spec modification for autoscaling

marians:
Based on and to be merged into https://github.com/giantswarm/api-spec/pull/121

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -152,11 +152,6 @@ definitions:
             type: integer
             description: |
               The maximum number of cluster nodes
-      workers:
-        type: array
-        description: Worker node array
-        items:
-          $ref: '#/definitions/V4NodeDefinition'
 
   # Details on existing cluster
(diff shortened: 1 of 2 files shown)


marians:
@xh3b4sd @teemow This is my update to the `workers` spec. If you like, we can merge this into https://github.com/giantswarm/api-spec/pull/121

xh3b4sd:
When we change the behaviour for workers we introduce a breaking change and create more work for ourselves that will be thrown away later on anyway. I thought we could simply keep the workers field as it is and keep using it as it works now in case it is given for an update. I thought when instead of the workers the scaling settings are given we use the new way. Like a switch. That way the API does not break, but is extended. We can deprecated more gracefully and move forward without unnecessary efforts.

teemow:
@xh3b4sd the worker array is kept like it is. only the number of elements doesn't have any effect anymore.

PR #124, opened 2019-01-09
Fill V4GetClusterStatusResponse with flesh

tuommaki:
Instead of defining free-form object, define the actual object as it is in
custom resource specification.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -240,7 +240,129 @@ definitions:
   V4GetClusterStatusResponse:
     type: object
     description: Object about a cluster's current state
-    additionalProperties: true
+    properties:
+      aws:
+        type: object
+        properties:
+          availabilityZones:
+            type: array
+            description: Array of availability zones used for this cluster.
+            items:
+              type: object
+              properties:
+                name:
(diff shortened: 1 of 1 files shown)


tuommaki:
I still don't understand why free-form object doesn't work, but after testing, this does (and all the other code in Happa is unmodified). ¯\_(ツ)_/¯

marians:
See this thread for the current state of the discussion: https://gigantic.slack.com/archives/C13MG9F0W/p1547041500003700

definitions.yaml
```diff
@@ -240,7 +240,129 @@ definitions:
...
+            items:
+              type: object
+              properties:
+                lastTransitionTime:
+                  type: string
+                  description: |
+                    lastTransitionTime is the last time the condition
+                    transitioned from one status to another.
+                status:
+                  type: string
```
```suggestion
                  type: string
                  description: Status to which has transited
```

tuommaki:
After discussion in #sig-ux channel (the above referenced thread) we decided to drop this and use separate client to get the proper response as such. Closing.

PR #125, opened 2019-01-10
Add info regarding scaling min=max

marians:
This adds an info how a missing `scalin.min` or `scaling.max` value will be treated.

From https://gigantic.slack.com/archives/C7865GLSY/p1547111558000800

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -110,7 +110,8 @@ definitions:
         type: object
         description: |
           Attributes specific to cluster node scaling. To have full control of
-          the cluster size, min and max can be set to the same value.
+          the cluster size, min and max can be set to the same value. If only
+          `min` or only `max` is specified, `min` and `max` will be set equally.
         properties:
           min:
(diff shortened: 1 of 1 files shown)


MarcelMue:
I didn't read this carefully enough, this also implies that we can't set the min workers to `0`.

marians:
Should it be possible to set min workers to 0?

teemow:
The autoscaler supports that. uncertain if this will really work atm. afaik we flagged some of the daemonsets to not be accounted but there are more things running by ourselves on the cluster and that could block the downscaling.

marians:
So if it's not expected to work, we should ideally block the attempt to set min=0 with a 400 error and also add an according info to the spec.

teemow:
It will work at some point. So please dont block this via the API. It is currently not the main point to make this work. Eg. if we block this via the API there is no chance for a solution engineer to make it work in certain clusters if the first use case is there.

PR #128, opened 2019-02-28
Add region field for cluster

tuommaki:
In order to be able to create inter-region clusters in AWS, a region field must
be present for cluster object.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -103,6 +103,9 @@ definitions:
         description: |
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           to use in the new cluster
+      region:
+        description: Region where tenant cluster should be created. The default is the same region where Giant Swarm control plane is running.
+        type: string
       availability_zones:
         description: Number of availability zones a cluster should be spread across. The default is provided via the [info](#operation/getInfo) endpoint.
         type: integer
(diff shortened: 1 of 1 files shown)


tuommaki:
This is not optimal, but aligns with `availability_zones`.

definitions.yaml
```diff
@@ -206,6 +209,9 @@ definitions:
             type: integer
             description: |
               The maximum number of cluster nodes as configured
+      region:
+        type: string
+        description: Region where this cluster is running at.
```
```suggestion
        description: Name of the region where this cluster is running at.
```
A bit more aligned with AZ and others

definitions.yaml
```diff
@@ -103,6 +103,9 @@ definitions:
         description: |
           The [release](https://docs.giantswarm.io/api/#tag/releases) version
           to use in the new cluster
+      region:
+        description: Region where tenant cluster should be created. The default is the same region where Giant Swarm control plane is running.
```
```suggestion
        description: Region where tenant cluster should be created. The default is the same region the Giant Swarm control plane is running in.
```

tuommaki:
Oh man, I forgot this. I'll close this PR for now as we discussed with Timo that we won't expose this just yet. We'll add the functionality and continue improving it, but we won't allow it to be used publicly before we have improved all necessary pieces.

PR #130, opened 2019-03-12
Node pools

marians:
Introducing capabilities to deal with *node pools*.

### What are node pools?

As of a future release version (version number to be defined), worker nodes in tenant clusters will be grouped into node pools. All nodes in a node pool will share the same configuration and scaling will be controlled on a node pool level. This will allow to fulfill a variety of different workload requirements within the same cluster.

Node pools will be fully supported first on **AWS only**. Here, clusters can then consist of multiple node pools. Node pools can be added, modified and deleted during the lifetime of a cluster.

### Info regarding breaking changes

This change does not introduce breaking changes.

We add a new version path `/v5` to prevent changes to existing API operations. The cluster resource for clusters featuring node pools will be found under the `/v5` path, while clusters using the old structure are still accessed via the `/v4` path.

### Hidden API operations

While the backend does not support node pools yet, the according API operations are hidden in the documentation. However they will be supported in generated clients.


teemow:
> Is there a node pool status resoure, similar to the current cluster status resource, to be added?

The machine deployment will have a status yes. And we should make this available similarly to how we did this with clusters before. The cluster status is also likely to be changed with this story because of Cluster API.

See:
 * https://kubernetes-sigs.github.io/cluster-api/common_code/cluster_controller.html#clusterstatus
 * https://kubernetes-sigs.github.io/cluster-api/common_code/machinedeployment_controller.html#machinedeploymentstatus

spec/definitions.yaml
```diff
@@ -962,6 +962,34 @@ definitions:
...
+      conditions:
+        type: array
+        description: List of conditions the cluster has gone through
+        items:
+          type: object
+          properties:
+            last_transition_time:
+              type: string
+              description: Date and time when the cluster transitioned into this condition
+            type:
```
xh3b4sd: When the list is called `conditions`, would it make sense to call this here `condition` instead? Below we have the list called `versions` and the item's field called `version`. That is consistent and comprehensible. I always struggled with `type` and now that we cleanup and revamp the status structures with new types, I would like to have more feedback about this and maybe make it better. I would then also like to adapt internally accordingly.

marians: I think you are right, `condition` would be better than `type`.

xh3b4sd: Let me know what the consensus is. Then I am happy to provide PRs for the internal changes in the operators and `apiextensions`.

marians: The consensus is `condition`. See https://github.com/giantswarm/api-spec/pull/130/commits/dc7a27c166ee9f3d0f0f6fb8ab111c0589bc19bf

spec/definitions.yaml
```diff
@@ -788,3 +795,268 @@ definitions:
...
+        type: object
+        $ref: '#/definitions/V5GetNodePoolResponseNodeSpec'
+
+
+  # Request to create a new cluster in V5
+  V5AddClusterRequest:
+    type: object
+    required:
+      - owner
+    properties:
```
tobbbles: Do we yet know which fields are required for creation?

spec/definitions.yaml
```diff
@@ -788,3 +795,268 @@ definitions:
...
+    type: array
+    items:
+      $ref: '#/definitions/V5GetNodePoolResponse'
+
+  # Add node pool
+  V5AddNodePoolRequest:
+    description: |
+      Request body structure for creating a node pool
+    type: object
+    properties:
```
tobbbles: Similary do we know which fields are required when creating a nodepool creation request?

spec/definitions.yaml
marians: Ideally none should be required and we would have a sensible default for any of them.

PR #131, opened 2019-04-05
List appcatalogs

oponder:
This adds a way to list all the `appcatalogs`. Basically a way for `happa` and other clients to do `kubectl get appcatalogs`. 

Towards: https://github.com/giantswarm/giantswarm/issues/4848


puja108:
But doesn't get appcatalogs usually just give back the list without details? In ok though if we make it a single call in GS API as the info is minimal and we're used to giving back more details per object on list calls

oponder:
Kubernetes actually gives me the full object, including `spec`:

```
$ kubectl get appcatalogs -o yaml
apiVersion: v1
items:
- apiVersion: application.giantswarm.io/v1alpha1
  kind: AppCatalog
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"application.giantswarm.io/v1alpha1","kind":"AppCatalog","metadata":{"annotations":{},"creationTimestamp":"2019-04-01T16:14:05Z","finalizers":["operatorkit.giantswarm.io/app-operator"],"generation":1,"labels":{"app-operator.giantswarm.io/version":"1.0.0"},"name":"sample-catalog","resourceVersion":"111672656","selfLink":"/apis/application.giantswarm.io/v1alpha1/appcatalogs/sample-catalog","uid":"2c28029d-5499-11e9-8be9-0aed6f318a48"},"spec":{"description":"Sample Catalog","storage":{"URL":"https://giantswarm.github.com/sample-catalog/","type":"helm"},"title":"Sample Catalog"}}
    creationTimestamp: "2019-04-05T05:25:21Z"
    finalizers:
    - operatorkit.giantswarm.io/app-operator
    generation: 1
    labels:
      app-operator.giantswarm.io/version: 1.0.0
    name: sample-catalog
    resourceVersion: "4821"
    selfLink: /apis/application.giantswarm.io/v1alpha1/appcatalogs/sample-catalog
    uid: 35290c03-5763-11e9-85bd-0242ac130006
  spec:
    description: Sample Catalog
    storage:
      URL: https://giantswarm.github.com/sample-catalog/
      type: helm
    title: Sample Catalog
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

puja108:
True -o yaml does that. Anyway, let's go with this

pipo02mix:
IMO return the list of entire appcatalogs is right, otherwise, how will you get all info? 
I think if you want just the name I would add a filter `fields` with the desired properties

definitions.yaml
```diff
@@ -12,6 +12,36 @@ definitions:
...
+  V4AppCatalogsResponse:
+    type: array
+    items:
+      type: object
+      properties:
+        description:
+          description: A description of the catalog.
+          type: string
+        logoURL:
+          description: A URL to a logo representing this catalog.
```
```suggestion
          description: An URL to a logo representing this catalog.
```
nitpick

spec.yaml
```diff
@@ -82,6 +82,58 @@ security:
...
 
 paths:
+  /v4/appcatalogs/:
+    get:
+      operationId: getAppCatalogs
+      tags:
+        - App Katalog
+      summary: Get a list of app catalogs configured on your installation.
+      description: |
+        Returns an array of app catalog objects, which contain further metadata,
```
pipo02mix:
```suggestion
        Returns an array of app catalog objects, which contains further metadata,
```
another nitpick

spec.yaml
```diff
@@ -82,6 +82,58 @@ security:
...
+        including a URL to fetch the full index of each catalog.
```
pipo02mix:
```suggestion
        including an URL to fetch the full index of each catalog.
```

definitions.yaml
oponder: I think "A URL" is correct?

I realise I didn't apply this consistently later in the definition though. So will change it all to `a url`

https://www.a-or-an.com/a_an/url

pipo02mix: ok good

spec.yaml
```diff
@@ -82,6 +82,69 @@ security:
   - AuthorizationHeaderToken: []
 
 paths:
+  /v4/appcatalogs/:
+    get:
+      operationId: getAppCatalogs
+      tags:
+        - app katalog
```
tomahawk28: Do we have a reason that sometimes we say it `Katalog`? (German?)

oponder: It's to distinguish our offering on the market, so `App Katalog` is kind of the brand name of this feature, which makes it recognisable as something from Giant Swarm.

PR #132, opened 2019-04-05
List tenant cluster apps.

oponder:
This PR proposes a way to fetch the apps that are installed on a given tenant cluster.

FYI @giantswarm/sig-ux : Operations can have multiple tags, so it will show up in the `Cluster` category as well as the `App Katalog` category in api docs.

That seems desirable to me since I would personally expect to see it in both places. But opinions are welcome here.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -284,6 +284,52 @@ definitions:
         type: object
         additionalProperties: true
 
+  V4GetClusterAppsResponse:
+    type: array
+    description: Array of apps
+    items:
+      type: object
+      properties:
+        metadata:
+          type: object
+          properties:
+            name:
+              type: string
+              description: The identifier you set when creating this app
+        spec:
+          type: object
+          properties:
+            name:
+              type: string
+              description: The name of the chart that was used to install this app
+            namespace:
+              type: string
+              description: The namespace that this app is installed to
+            version:
+              type: string
+              description: The version of the chart that was used to install this app
+            catalog:
+              type: string
+              description: The catalog that this app came from
+        status:
+          type: object
+          properties:
+            app_version:
+              type: string
+              description: The version of the installed app
+            version:
+              type: string
+              description: The version of the chart that was used to install this app
+            release:
+              type: object
+              properties:
+                last_deployed:
+                  type: string
+                  description: Date and time that the app was last last deployed
+                status:
+                  type: string
+                  description: "A string representing the status of the app. (Can be: empty, DEPLOYED or FAILED)"
+
   V4GetClusterStatusResponse:
     type: object
     description: Object about a cluster's current state
--- a/spec.yaml
+++ b/spec.yaml
@@ -1088,6 +1088,14 @@ paths:
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v4/clusters/{cluster_id}/apps/:
+    get:
+      operationId: getClusterApps
+      tags:
+        - app katalog
+        - cluster
+      summary: Get cluster apps
+
   /v4/organizations/:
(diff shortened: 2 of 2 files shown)


spec.yaml
```diff
@@ -1088,6 +1088,85 @@ paths:
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v4/clusters/{cluster_id}/apps/:
+    get:
+      operationId: getClusterApps
+      tags:
+        - app katalog
+        - clusters
```
oponder: this means it shows up both in the `Clusters` and `App Katalog` tab in the api docs.

pipo02mix: hmm but it is only returning installed apps on that cluster, or the list of all apps? (installed and available to install)

oponder: Only the apps installed on that cluster.

pipo02mix: then I am not sure it will be correct list the endpoint on the app katalog. I see it can get mixed with the list available apps, dont you think?

oponder: hmm.. would like some other feedback on that. To me its pretty clear.

list of available apps is the helm repo index at the `url` you get for a catalog when doing 

`GET /v4/appcatalogs/`  (different PR)

and then the list of apps on a cluster is

`GET /v4/clusters/:clusterid/apps/`  (this pr)

definitions.yaml
```diff
@@ -284,6 +284,52 @@ definitions:
...
+    description: Array of apps
+    items:
+      type: object
+      properties:
+        metadata:
+          type: object
+          properties:
+            name:
+              type: string
+              description: The identifier you set when creating this app
```
puja108: I don't think we have description in app CRD, it's a desire to install an app, you don't need to describe the app for that

spec.yaml
puja108: yeah, I think Fer is wondering if listing this under catalog would make people think there /ve/appcatalogs/:catalog/apps

definitions.yaml
puja108: nvm, I'm stupid ;)

oponder: I got confused for a second there too :P

oponder:
> let's think about the linking and if it least to confusion

For me it's fine, it might not be exactly what you expect, but then you read it and it makes sense.

At least you'll find it in the places where you might expect to find it right?

puja108:
ack, go with it

oponder:
@marians: There are two resources that we currently want to expose or have access to. 

Those resources are `apps` (on a tenant cluster) and `appcatalogs` (an installation wide concern).

Apps might have subresources that we want to configure (like secrets or configmaps) but it's not decided yet if that'll just be part of the `app` resource or if they will be separate sub resources.

I can imagine the following operations and endpoints then:

```
Get all catalogs configured for this installation:
  GET  /v4/appcatalogs/

Add a catalog configured for this installation:
  POST /v4/appcatalogs/

Get all apps installed on a tenant cluster:
  GET  /v4/clusters/:clusterid/apps/

Get more details about an installed app:
  GET  /v4/clusters/:clusterid/apps/:appname/

Install an app:
  PUT /v4/clusters/:clusterid/apps/:appname/

Delete an app:
  DELETE /v4/clusters/:clusterid/apps/:appname/

Make changes to an app:
  PATCH /v4/clusters/:clusterid/apps/:appname/
```

oponder:
I'll be moving forward with this spec and start implementing `GET /v4/clusters/:clusterid/apps/` now.

ping @marians , any comments since you requested more info before?

marians:
Fine with proceeding here! Thanks for providing more context. That helps!

PR #135, opened 2019-05-02
Add the managed apps topic.

oponder:
This PR takes the accepted specs so far for the managed apps topic and groups in one place.

`GET /v4/appcatalogs/` https://github.com/giantswarm/api-spec/pull/131
`GET /v4/clusters/:clusterid/apps/` https://github.com/giantswarm/api-spec/pull/132
`PUT /v4/clusters/:clusterid/apps/:appname/` https://github.com/giantswarm/api-spec/pull/136
`DELETE /v4/clusters/:clusterid/apps/:appname/` https://github.com/giantswarm/api-spec/pull/137

Towards: https://github.com/giantswarm/giantswarm/issues/4848

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -1231,6 +1231,49 @@ paths:
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v4/clusters/{cluster_id}/apps/{app_name}/:
+    delete:
+      operationId: deleteClusterApp
+      tags:
+        - managed apps
+      summary: Delete an app
+      description: |
+        This operation allows a user to delete an app.
+      parameters:
+        - $ref: './parameters.yaml#/parameters/RequiredGiantSwarmAuthorizationHeader'
+        - $ref: './parameters.yaml#/parameters/XRequestIDHeader'
+        - $ref: './parameters.yaml#/parameters/XGiantSwarmActivityHeader'
+        - $ref: './parameters.yaml#/parameters/XGiantSwarmCmdLineHeader'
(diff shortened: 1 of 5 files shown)


spec.yaml
```diff
@@ -43,6 +43,8 @@ tags:
     externalDocs:
       url: https://docs.giantswarm.io/guides/accessing-services-from-the-outside/
       description: "User guide: Accessing Pods and Services from the Outside"
+  - name: managed apps
+    description: "An app is a helm chart that you can easily install on your tenant clusters. Use these endpoints to: find out what catalogs your installation supports, and list and install apps on on your tenant clusters."
```
marians: In terms of the topic/tag here, I would suggest to call it `apps` instead of `managed apps`.

Terminology in routes like `/v4/clusters/{cluster_id}/apps/` suggests that we call those things "apps". The fact that they are sometimes managed should not be of importance for the general grouping in the API spec.

oponder: That makes sense.

Right now we're grouping also the `/v4/appcatalogs/` call into this `managed apps` tag.

The idea was to brand these set of endpoints as a feature, the `managed apps` feature.

That said, in the api docs, it might be fine if we introduce two topics. `apps` and `appcatalog`.

What do you think?

marians: If app catalogs exist purely to allow installing apps, then I think it would be fine to group it all under the topic of `apps`. But no strong opinion on this one now.

Let's try whatever you find useful and see if it makes sense in ReDoc.

oponder: Alright, I'll try grouping it all under `apps`. Will hold of on that change until the installing apps PR is merged into this one. Then I make the change all in one go here.

PR #136, opened 2019-05-06
Installing apps

oponder:
This specs the `POST /v4/clusters/{clusterid}/apps/` call.

The idea is that the user posts a representation of the App CR as they want to have it, and the API will create it in the right namespace on the control plane based on the clusterid parameter in the path.

A sample post request:

```
          {
            "metadata": {
              "name": "my-awesome-prometheus",
            },
            "spec": {
              "catalog": "sample-catalog",
              "name": "prometheus-chart",
              "namespace": "prometheus",
              "version": "0.2.0"
            }
          }
```

Because we return the a representation of the created resource, we can extract the object from the array response of `V4GetClusterAppsResponse` and use it in both this POST call and the GET call.

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -315,51 +315,54 @@ definitions:
         type: object
         additionalProperties: true
 
+  V4App:
+    type: object
+    properties:
+      metadata:
+        type: object
+        properties:
+          name:
+            type: string
+            description: The identifier you set when creating this app
+      spec:
+        type: object
+        properties:
+          name:
+            type: string
+            description: The name of the chart that was used to install this app
+          namespace:
+            type: string
+            description: The namespace that this app is installed to
+          version:
+            type: string
+            description: The version of the chart that was used to install this app
+          catalog:
+            type: string
+            description: The catalog that this app came from
+      status:
+        type: object
+        properties:
+          app_version:
+            type: string
+            description: The version of the installed app
+          version:
+            type: string
+            description: The version of the chart that was used to install this app
+          release:
+            type: object
+            properties:
+              last_deployed:
+                type: string
+                description: Date and time that the app was last last deployed
+              status:
+                type: string
+                description: "A string representing the status of the app. (Can be: empty, DEPLOYED or FAILED)"
+
   V4GetClusterAppsResponse:
     type: array
     description: Array of apps
     items:
-      type: object
-      properties:
-        metadata:
-          type: object
-          properties:
-            name:
-              type: string
-              description: The identifier you set when creating this app
-        spec:
-          type: object
-          properties:
-            name:
-              type: string
-              description: The name of the chart that was used to install this app
-            namespace:
-              type: string
-              description: The namespace that this app is installed to
-            version:
-              type: string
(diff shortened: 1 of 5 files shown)


spec.yaml
```diff
@@ -1088,6 +1153,163 @@ paths:
...
+                    "last_deployed": "0000-00-00T00:00:00Z",
+                    "status": ""
+                  },
+                  "version": ""
+                }
+              }
+
+        "401":
+          $ref: "./responses.yaml#/responses/V4Generic401Response"
+        default:
```
pipo02mix: Do we need a "409 Conflict" to respond when app already exists?

oponder: Hmm good point. Though I think I'd like to also go for a `400`, as a general user input validation error.

pipo02mix: ok good, no strong feelings about the code type

pipo02mix: but do we have this as default? 🤔

definitions.yaml
```diff
@@ -315,51 +315,54 @@ definitions:
...
+        properties:
+          name:
+            type: string
+            description: The identifier you set when creating this app
+      spec:
+        type: object
+        properties:
+          name:
+            type: string
+            description: The name of the chart that was used to install this app
```
```suggestion
            description: Name of the chart that was used to install this app
```

definitions.yaml
```diff
@@ -315,51 +315,54 @@ definitions:
...
+          namespace:
+            type: string
+            description: The namespace that this app is installed to
```
```suggestion
            description: Namespace that this app is installed to
```

definitions.yaml
```diff
@@ -315,51 +315,54 @@ definitions:
...
+          version:
+            type: string
+            description: The version of the chart that was used to install this app
```
```suggestion
            description: Version of the chart that was used to install this app
```

definitions.yaml
```diff
@@ -315,51 +315,54 @@ definitions:
...
+          catalog:
+            type: string
+            description: The catalog that this app came from
+      status:
+        type: object
+        properties:
+          app_version:
+            type: string
+            description: The version of the installed app
```
```suggestion
            description: Version of the installed app
```

definitions.yaml
```diff
@@ -315,51 +315,54 @@ definitions:
...
+          version:
+            type: string
+            description: The version of the chart that was used to install this app
```
```suggestion
            description: Version of the chart that was used to install this app
```

marians:
The descriptions stuttered a bit for me. Added some suggestions to remove some `The`.

spec.yaml
oponder: > but do we have this as default?

Do you mean do we have some conventions around this? We use `409` indeed when something already exists.

But what I mean is I think I'll add both `409` and `400`.

oponder:
@marians quick question, do you prefer 

`PUT /v4/clusters/{clusterid}/apps/{appname}/` 

instead of 

`POST /v4/clusters/{clusterid}/apps/` (as it is specced now)

Since we do that in cases where the identifier is known by the user before creation.
See org creation for example.

marians:
@oponder Yes, in fact, if the path is known, then we should use PUT.

PR #137, opened 2019-05-06
Add spec for deleting apps.

oponder:
Relatively straight forward, this adds a way to delete an app at `DELETE /v4/clusters/{cluster_id}/apps/{app_name}/`

### DIFF
--- a/parameters.yaml
+++ b/parameters.yaml
@@ -8,6 +8,13 @@ parameters:
     description: |
       As described in the [authentication](#section/Authentication) section
 
+  AppNamePathParameter:
+    name: app_name
+    in: path
+    required: true
+    type: string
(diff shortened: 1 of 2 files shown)


marians:
Looks good!

Who defines `app_name` and when/how/where? Is it done by the user on installation? Let's restrict the app name to a very small set of characters (a-z, 0-9, _, -) without blanks, please.

oponder:
> Let's restrict the app name to a very small set of characters (a-z, 0-9, _, -) without blanks, please.

This will also already be restricted by kubernetes, since it is the `metadata.name` field of the App CR.

But I will add validation for that in the Giant Swarm API when creating an app as well. And then in Happa client side for the convenience of the user.

PR #145, opened 2019-05-17
Create app config.

oponder:
This specs the `PUT /v4/clusters/{cluster_id}/apps/{app_name}/config/` endpoint

~This would be the first endpoint takes a file as its parameter instead of JSON in the body.~

This endpoint takes JSON as normal.

Will be merged into: https://github.com/giantswarm/api-spec/pull/143

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -1382,7 +1382,7 @@ paths:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
   /v4/clusters/{cluster_id}/apps/{app_name}/config/:
-    post:
+    put:
       operationId: createClusterAppConfig
       tags:
         - apps
--- a/spec.yaml
+++ b/spec.yaml
@@ -1384,14 +1384,27 @@ paths:
   /v4/clusters/{cluster_id}/apps/{app_name}/config/:
     post:
       operationId: createClusterAppConfig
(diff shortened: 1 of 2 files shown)


spec.yaml
```diff
@@ -1381,6 +1381,78 @@ paths:
...
+    post:
+      operationId: createClusterAppConfig
+      consumes:
+        - multipart/form-data
+      tags:
+        - apps
+      summary: Create app config
+      description: |
+        This operation allows you to create a values configmap for a specific app. The app does
+        have to exist before hand.
```
pipo02mix: It sounds weird you are creating a subresource of resource that does not exist. Why do we want to do it like that?
Can be a configmap use for multiple appCRs (same type)?

oponder: > Can be a configmap use for multiple appCRs

Via the kubernetes API (App CR spec) yes, via the GS API, no.

The App CR spec has an explicit reference to the configmap used for the user values, so anything can go there.

For the case of the API, it's about setting up a convention so that we can get a configmap in place before making the App CR, so that the App gets installed properly in one go.

oponder:
FYI I am changing this from `POST` to `PUT`, to be in line with our convention. 

This is a named resource, so it should be `PUT` right.

PR #147, opened 2019-06-05
Add user_config to app creation and listing.

oponder:
This adds a way for API users to associate their app with one of the configmaps they can create using the endpoints we're introducing here: https://github.com/giantswarm/api-spec/pull/143

### DIFF
--- a/definitions.yaml
+++ b/definitions.yaml
@@ -360,6 +360,8 @@ definitions:
           catalog:
             type: string
             description: The catalog where the chart for this app can be found
+          user_config:
+            $ref: '#/definitions/V4AppSpecUserConfigConfigMap'
 
   V4App:
     type: object
@@ -385,6 +387,9 @@ definitions:
           catalog:
             type: string
             description: The catalog that this app came from
+          user_config:
+            $ref: '#/definitions/V4AppSpecUserConfigConfigMap'
+
       status:
         type: object
         properties:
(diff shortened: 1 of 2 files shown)


marians:
The name `V4AppSpecUserConfigConfigMap` is stuttering. Could we make it `V4AppSpecUserConfigMap` instead?

definitions.yaml
```diff
@@ -385,6 +387,9 @@ definitions:
           catalog:
             type: string
             description: The catalog that this app came from
+          user_config:
+            $ref: '#/definitions/V4AppSpecUserConfigConfigMap'
+
```
marians: This empty line could be removed

definitions.yaml
```diff
@@ -404,6 +409,20 @@ definitions:
                 type: string
                 description: "A string representing the status of the app. (Can be: empty, DEPLOYED or FAILED)"
 
+  V4AppSpecUserConfigConfigMap:
+    type: object
+    description: User definable configuration to be applied when the app is deployed
+    properties:
+      config_map:
```
marians: The name `config_map` suggests that this represents an entire Kubernetes configmap. My understanding was that this only represents the `data.values` part. So maybe name it `configmap_values` or `configmap_data_values`?

marians: Oh, strike that. I was in a different thread mentally.

As this seems to reference a `ConfigMap` name, in general the name is fine. My only issue is that I'd rather translate it to `configmap` than `config_map`.

marians:
Have you considered this simpler alternative?

```json
"user_configmap": {
  "name": "prometheus-user-values",
  "namespace": "123ab"
}
```

oponder:
> Have you considered this simpler alternative?
> 
> "user_configmap": {
>   "name": "prometheus-user-values",
>   "namespace": "123ab"
> }

In this case, I am trying to mirror the CR as closely as possible. 

Though I am fine with changing `config_map` to `configmap`

PR #149, opened 2019-06-19
Clean things up a bit

oponder:
I'm preparing for a move to OpenAPI 3.0

I'd like to get the actual spec files in their own folder, so I started doing that, and noticed a few things along the way.

- Move all actual spec files into their own folder
- Don't create temporary files
- Don't run linter twice in circleci
- Don't output the temporary javascript client back to the host system


marians:
Nice in general! Does the spec format change somehow? It's a bit unfortunate that the diff doesn't allow reviewing that, as the files are also moved with the same commit.

We should also prepare to update the following three repos:

- docs-indexer: The spec is parsed using `openapi-spec-validator` to allow specific search for endpoints.
- gsclientgen: Makefile needs path adjustment here https://github.com/giantswarm/gsclientgen/blob/master/Makefile#L11-L14 - also, is OpenAPI 3 supported by the generator?
- Similar for the JS client here: https://github.com/giantswarm/giantswarm-js-client/blob/master/Makefile#L30-L33

oponder:
@marians no change in the spec, just moving files.

OpenAPI 3 change is more involved and will be its own PR.

oponder:
- gsclientgen: https://github.com/giantswarm/gsclientgen/pull/44
- giantswarm-js-client: https://github.com/giantswarm/giantswarm-js-client/pull/68
- docs-indexer: that repo does parsing of the live site at: https://docs.giantswarm.io/api/ so no changes needed.

marians:
Thanks for also doing the follow-up chores!

PR #150, opened 2019-06-20
Add spec for deleting app config.

oponder:
To be merged into: https://github.com/giantswarm/api-spec/pull/143

This allows clients to delete app configs if they follow the naming convention of app-name-user-values.

The configs that Happa and eventually gsctl, creates will always follow this convention, since it is built into the API.

Clients are able to discover if the user_config configmap is following the convention, and so indicate whether or not they are able to delete the configmap or not, and perhaps offer instructions on how to change the configmap (when users get access to the k8s api of the control plane.)

I went with this approach, because the path we've implement is very convention based.

Otherwise I'd have to fetch the App CR, check for the configmap reference, delete the configmap, and then go back to the App CR and unset the reference? I think at that point the API might be doing too much. Also non of the other endpoints implemented so far treat the App CR and config as one thing that gets updated at the same time. So for example, when creating a config, it's not like that endpoint automatically updates the App CR and sets the reference to the newly created config.


spec.yaml
```diff
@@ -1402,6 +1432,184 @@ paths:
...
+                  "endpointPort": "443",
+                },
+                "zone": {
+                  "name": "giantswarm-cluster"
+                }
+              }
+        "401":
+          $ref: "./responses.yaml#/responses/V4Generic401Response"
+        "404":
+          description: App not found
```
pipo02mix:
```suggestion
          description: App Config not found
```

PR #151, opened 2019-06-25
Add spec for modifying apps and app configs.

oponder:
This is a spec for the PATCH requests on the `app` and `app config` resources.

I made it so only the `user_config` and `version` fields of `app` are editable. I feel like changing the catalog, name, or namespace would probably be complex, not sure how the operators would react to that.

For the `app config`, as long as the configmap is named in the conventional way, users can change anything, using JSON merge patch semantics.

### DIFF
--- a/spec/spec.yaml
+++ b/spec/spec.yaml
@@ -45,6 +45,8 @@ tags:
       description: "User guide: Accessing Pods and Services from the Outside"
   - name: apps
     description: "An app is a helm chart that you can easily install on your tenant clusters. Use these endpoints to: find out what catalogs your installation supports, and list and install apps on on your tenant clusters."
+  - name: app configs
+    description: "Creating an app config lets you supply your apps with custom configuration values."
   - name: organizations
     description: Organizations are groups of users who own resources like clusters.
   - name: users
@@ -1436,7 +1438,7 @@ paths:
       operationId: modifyClusterApp
       tags:
         - apps
-      summary: Modify app
+      summary: Modify an app
       description: |
         This operation allows you to modify an existing app.
 
@@ -1444,7 +1446,7 @@ paths:
 
         - `version`: Changing this field lets you upgrade or downgrade an app.
 
-        `catalog`, `name`, `namespace`, and 'user_config' are not editable. If you need to move or rename an app, you should instead delete the app and make it again.
+        `catalog`, `name`, `namespace`, and `user_config` are not editable. If you need to move or rename an app, you should instead delete the app and make it again.
 
         The request body must conform with the [JSON Patch Merge (RFC 7386)](https://tools.ietf.org/html/rfc7386) standard.
         Requests have to be sent with the `Content-Type: application/merge-patch+json` header.
@@ -1502,7 +1504,7 @@ paths:
     get:
       operationId: getClusterAppConfig
       tags:
-        - apps
+        - app configs
       summary: Get app config
       description: |
         This operation allows you to fetch the user values configmap associated
@@ -1546,7 +1548,7 @@ paths:
     put:
       operationId: createClusterAppConfig
       tags:
-        - apps
+        - app configs
       summary: Create app config
       description: |
         This operation allows you to create a values configmap for a specific app. The app does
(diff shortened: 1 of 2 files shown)


oponder:
pinging @rossf7 specifically to ask for some input on the editing of `App CR` behaviour and if my assumptions that only allowing to edit version and user_config here make sense.

rossf7:
@oponder For the app CRs I'm not sure changing any fields except `version` makes sense.

AIUI we want to default the config fields using OPA. e.g. For CoreDNS the user CM is always called `coredns-user-values`. So I don't think changing the name of the user CM makes sense.

@puja108 WDYT?

For the user CM structure, it needs to have a single key and the value has to be YAML. The name of the key doesn't matter but for the generated CMs we use `values`.

```
Data: map[string]string{
	"values": "baseDomain: giantswarm.io\n",
},
```

These YAML values are merged with the rest of the values.

spec/spec.yaml
```diff
@@ -1432,6 +1432,73 @@ paths:
...
+      operationId: modifyClusterApp
+      tags:
+        - apps
+      summary: Modify app
+      description: |
+        This operation allows you to modify an existing app.
+
+        The following attributes can be modified:
+
+        - `user_config`: This lets you change which configmap is used to configure this app, as well as remove it entirely, by submitting null values.
```
rossf7: Removing the config is a good use case but then the OPA code might add it again? I think removing the values from the user CM is a better solution.

oponder:
> For the user CM structure, it needs to have a single key and the value has to be YAML. The name of the key doesn't matter but for the generated CMs we use values.

@rossf7 : right, the API adds what gets sent as body json to the `values` field of the configmap. So the user doesn't have to remember to add the values key

rossf7:
@oponder Cool that's great. :+1:

spec/spec.yaml
oponder: So also when creating an app, there is no real need to make the user_config something that users can set themselves via the API right?

oponder:
Also decided to move the App Config endpoints to their own 'tag' and so into their own category from the side menu:

Screenshot 2019-07-04 at 9 06 07 PM

spec/spec.yaml
```diff
@@ -1611,6 +1679,74 @@ paths:
...
+              }
+        "400":
+          description: Invalid input
+          schema:
+            $ref: "./definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "INVALID_INPUT",
+                "message": "App config could not be created. (invalid input)"
```
pipo02mix:
```suggestion
                "message": "App config could not be modified. (invalid input)"
```

spec/spec.yaml
puja108: I would like to support removing, and I do not think OPA should add user configs, OPA should only fill in cluster config

puja108:
I would say cahnging configmap is legitimate and we should not use OPA to default configmap names, I specifically want users to be able to have configmaps they can share between clusters and defaulting does not work there. Limiting to version and user config is ok though, I don't see any other need for editing right now. /cc @rossf7

spec/spec.yaml
rossf7: Ack. That makes sense to me.

oponder:
After some discussion and realisation, we decided it would be best if the API took care of setting the `user_config` field correctly for the user, depending on what is going on.

So the `user_config` field will not be something that the user can set. They can still see it though in the get calls and in the response to the create call to verify it is set correctly.

So if the user creates a config, then the API will also edit the App CR and make sure it points to the right thing.

And if the user creates an app, and a correctly named configmap already exists on the control plane, then the API will make sure it gets referenced.

I added the following text at the app creation endpoint to explain this:

>### About the user_config field in the response
>This field is not editable by you, but is set automatically by the API if a configmap named {app_name}-user-values exists in the tenant cluster namespace on the control plane.

>The /v4/clusters/{cluster_id}/apps/{app_name}/config/ endpoints allows you to create such a configmap using this API.

> It is recommended to create your config before creating your app. This will result in a faster deploy.

> However, you can create your config after creating the app if you wish, this API will take care of setting the user_config field of the app correctly for you.

>### Why can't I just set the user_config value myself?
>This is a security measure and ensures that users of this API can't access arbitrary configmaps of the control plane.

>This API will only allow you to edit or access configmaps that adhere to a strict naming convention.
```

PR #155, opened 2019-07-04
Modify Node Pool details

marians:
This PR changes #130 so that there is no separate node pools status endpoint.


marians:
I'd like to know whether the "subnet" term, coming from the annotation `machine-deployment.giantswarm.io/subnet: 10.1.0.0/24`, is named well and makes sense here. So far I have only guessed the meaning. @tuommaki Can you explain?

Regarding naming, I remember we used the term "CIDR" somewhere (can't find where).

tuommaki:
> I'd like to know whether the "subnet" term, coming from the annotation machine-deployment.giantswarm.io/subnet: 10.1.0.0/24, is named well and makes sense here. So far I have only guessed the meaning. @tuommaki Can you explain?

[*Subnet*](https://en.wikipedia.org/wiki/Subnetwork) is pretty much as it says: Sub net of some parent net. Smaller part of something bigger. In this context this basically means the network that was allocated for specific node pool (because each node pool gets its own piece of network, own piece of that cake I talked earlier today).

> Regarding naming, I remember we used the term "CIDR" somewhere (can't find where).

CIDR = [Classless Inter-Domain Routing](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) which basically is a term for specific way of expressing IP networks and using those expressions for routing purposes.

When we express a network in human readable format for anything, it's always in CIDR format (i.e. network address + size of the mask).


TL;DR; I'm not fully into what's best naming here, but `subnet` is not wrong. It could maybe also be `network`, but after thinking I'd maybe lean towards `subnet`. The object here is `Node Pool` and it's member variable is `subnet` so I guess that's ok. I'm open for discussion though. :)

marians:
Sounds good to me, thanks!

PR #156, opened 2019-07-09
Node Pool Creation Validation

tobbbles:
Fields to be validated
---

| Field | Optional | Validation |
| ---- | --- | --- |
| NodePool Name | True | Length 0-100, block control characters |
| NodePool Scaling | True | `max >= min >= 1`. Default to 3 min max if both empty. Default to 3 min, if max provided > 3 |
| Availability Zones | - | Only one of `number` or `zones` |
| Availability Zones - `number` | True | `installations max >= number >= 1` (Check with Tuomas).  Default to 1 if no names nor numbers given |
| Availability Zones - `zones` | True | Regexp to validate name format `a-z, -, 0-9`. Validate against cluster region |
| Node Spec AWS Instance Type | True | Validate against info instance type list |

Invalid requests to return `400 Bad Request` with description of validation error

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -811,7 +811,7 @@ definitions:
     properties:
       name:
         description: |
-          Node pool name (max character length of x)
+          Node pool name. _(Length between 1-100, cannot contain control codes such as newline.)_
         type: string
       availability_zones:
         description: |
@@ -825,11 +825,11 @@ definitions:
             type: integer
             description: |
               Number of zones to use. If given, the zones are picked
-              automatically.
+              automatically. _(Maximum limit of 4 supported.)_
           zones:
             type: array
             description: |
-              Names of the availability zones to use.
+              Names of the availability zones to use. _(Must be same region as the cluster.)_
             items:
               type: string
       scaling:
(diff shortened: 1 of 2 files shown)


spec/definitions.yaml
```diff
@@ -840,10 +840,10 @@ definitions:
         type: object
         properties:
           min:
-            description: Minimum number of nodes in the pool
+            description: Minimum number of nodes in the pool, must be greater than 0, less than `max`.
```
marians: "less than max" is not exactly right. min can be equal to max. This is the setting that also disables autoscaling.

spec/definitions.yaml
```diff
@@ -840,10 +840,10 @@ definitions:
...
             type: integer
           max:
-            description: Maximum number of nodes in the pool
+            description: Maximum number of nodes in the pool, must be greater than `min`.
```
marians: see above

marians: Since the validation of min and max depends on each other, I think we could move this to the description of the `scaling` attribute. WDYT?

tobbbles: Having the description in the `scaling` part makes sense to me

PR #158, opened 2019-07-22
Hide new endpoints

marians:
This PR adds the capability to hide some endpoints which are not ready for use. It applies this to all node-pool specific endpoints.


marians:
Pinging SIG UX for info only

marians:
@tobbbles Yes, good point. Added a paragraph on that.

PR #159, opened 2019-07-29
Add catalog-type labels

oponder:
This adds the labels on an `app catalog CR` to the `get catalogs` api endpoint output.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -24,6 +24,14 @@ definitions:
             name:
               description: A URL friendly identifier for the catalog.
               type: string
+            labels:
+              description: The labels that are set on this app catalog
+              type: object
+              properties:
+                "application.giantswarm.io/catalog-type":
+                  type: string
+                "app-operator.giantswarm.io/version":
+                  type: string
         spec:
           type: object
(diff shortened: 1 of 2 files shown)


spec/definitions.yaml
```diff
@@ -24,6 +24,14 @@ definitions:
             name:
               description: A URL friendly identifier for the catalog.
               type: string
+            labels:
+              description: The labels that are set on this app catalog
+              type: object
+              properties:
+                "application.giantswarm.io/catalog-type":
+                  type: string
+                "app-operator.giantswarm.io/version":
```
oponder: Maybe this isn't needed to be exposed via the api... Wasn't sure about including it or not.

pipo02mix: If we will let users add catalog version I am ok to make it visible

rossf7: This label is used by app-operator to decide which version to execute. It's the replacement for the VersionBundle fields we have in legacy CRDs.

I don't think it does harm to show it. But eventually we will use OPA defaulting here because it's an implementation detail customers shouldn't need to care about.

PR #160, opened 2019-07-29
Update before checkout.

oponder:
There's a weird problem currently where this repo sometimes fails to build.

After some googling, I found that doing any apt-get steps first might be a way to mitigate this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,12 @@ jobs:
   build:
     machine: true
     steps:
+      - run:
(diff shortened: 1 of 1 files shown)


oponder:
Read more: https://unix.stackexchange.com/questions/315502/how-to-disable-apt-daily-service-on-ubuntu-cloud-vm-image/315517#315517

oponder:
And also: https://discuss.circleci.com/t/docker-convenience-images-updates-july-2019/31239

PR #162, opened 2019-08-06
Upgrade to open api v3

oponder:
Upgrades the spec to open api v3 and ensures the validating and linting are still meaningful.


oponder:
Hey @marians @ograu, here is my work so far on migrating to openapi v3.

I also split up the paths into individual files, which made this PR a lot larger in terms of file count, but technically still the same number of lines.

I'm keeping this as a draft for now, since I still need to do changes to the js client and go client, and make sure they play nicely with gsctl and happa. 

But I just wanted to get your eyes and opinions on this already.

If you have some time, please check out the repo and try `make run-server`

The docs are styled just very slightly differently:

Screenshot 2019-08-07 at 12 30 47 PM

oponder:
I paused this, see my comment here: https://github.com/giantswarm/giantswarm/issues/5992#issuecomment-522414613

> I'm really sad to say this, but I can't get the go client working and don't think it's worthwhile to invest further time into it for now.

> Perhaps when https://github.com/go-swagger/go-swagger supports openapi v3, this'll be a seamless switch to make.

> For now though, and I realised this too late, since we didn't use swagger-codegen (or openapitools/openapi-generator) the differences between the generated go clients are pretty big, and I don't have the appetite to do that refactoring in gsctl right now.

> My suggestion will be to park the work for now. Stick to swagger 2.0, but keep my 3.0 work for a time when go-swagger supports openapi v3.

PR #163, opened 2019-08-16
Add features section to info response

marians:
Towards https://github.com/giantswarm/giantswarm/issues/5465

This PR adds an attribute to the `GET /v4/info/` response that will allow clients to detect from which release version on support for node pools will be available.


marians:
I have moved the `features` key up one level.

marians:
FYI @giantswarm/sig-ux We are adding this section to the info endpoint to enable clients to understand from which release on the feature "node pools" will be available.

PR #164, opened 2019-08-21
Add secret CRUD endpoints to api-spec.

oponder:
This adds CRUD endpoints at `/v4/clusters/{cluster_id}/apps/{app_name}/secret/`

@rossf7 and others, are there any decisions made yet regarding the conventional name for the user config secret?

Like how the configmap should be called `{app-name}-user-values`.
What should / will the secret be called?

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -470,6 +470,16 @@ definitions:
     description: A part of the App CR that we create behind the scenes
     additionalProperties: true
 
+  V4GetClusterAppSecretResponse:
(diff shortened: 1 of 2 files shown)


oponder:
No worries ross! Your feedback was totally timely! I'll use `{app-name}-user-secrets` and update that here in the docs, as well as in the code of the api endpoints 👍 

Keeping this PR open for now while I work on the implementation. 

TODO:
- [ ] Merge once API endpoints are tested and live.

PR #166, opened 2019-09-05
Remove nodepools from addClusterV5 request body

marians:
Towards https://github.com/giantswarm/giantswarm/issues/5465

As detailed in https://github.com/giantswarm/giantswarm/issues/5465#issuecomment-528360513 we decided to simplify v5 cluster creation, removing the possibility to define node pools in the same request.


marians:
Ping @tuommaki - This needs your eyeballs

PR #167, opened 2019-09-14
Add minimum cpu and memory sizes to info endpoint.

oponder:
I've introduced a way to validate minimum CPU Core and Memory values on KVM installations. 

It would be nice to expose this information so that our clients can also act on it and set sensible starting values.

see: https://github.com/giantswarm/api/pull/855
see: https://github.com/giantswarm/installations/pull/887


oponder:
Closing, this is old and we don't wanna add new things to the API.

The clients will have to get this information in a different way, from the control plane, somehow.

PR #169, opened 2019-09-27
Add availability zone names to info endpoint response

marians:
Towards https://github.com/giantswarm/giantswarm/issues/7105

This PR adds the capability to let the info endpoint (`GET /v4/info/`) inform about the distinct availability zone names usable in a region.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -91,7 +91,7 @@ definitions:
                 type: integer
               zones:
                 type: array
-                description: The distinct availability zones available in the installation's region.
+                description: The availability zones available for in the installation's region for use with tenant clusters.
                 items:
                   type: string
       features:
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -91,7 +91,7 @@ definitions:
                 type: integer
               zones:
                 type: array
-                description: The availability zones available for in the installation's region for use with tenant clusters.
+                description: The availability zones available in the installation's region for use with tenant clusters.
                 items:
                   type: string
       features:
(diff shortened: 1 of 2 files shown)


spec/definitions.yaml
```diff
@@ -89,6 +89,11 @@ definitions:
               default:
                 description: Default number of availability zones for a cluster.
                 type: integer
+              zones:
+                type: array
+                description: The distinct availability zones available in the installation's region.
```
oponder: Is 'distinct' a necessary word here?

marians: Don't know. Calling some native speakers. @rossf7 @JosephSalisbury @tfussell Does it sound better without?

tuommaki:
```suggestion
                description: Availability zone options for tenant cluster.
```
How about this? I'd maybe drop the `installation` from there since it's actually for the tenant cluster and I can imagine that in some corner case we could even restrict by certain BYOC account. I don't know - in all cases it still applies to tenant cluster account, not the g8s control plane / installation, so that's why I'm slightly picky here.

marians: Changing to `The availability zones available for in the installation's region for use with tenant clusters.`

marians:
@tuommaki PTAL

spec/definitions.yaml
tuommaki: Maybe zap first `for`, but otherwise I'm good with that.

spec/definitions.yaml
```diff
@@ -89,6 +89,11 @@ definitions:
               default:
                 description: Default number of availability zones for a cluster.
                 type: integer
+              zones:
+                type: array
+                description: The availability zones available for in the installation's region for use with tenant clusters.
```
tuommaki:
```suggestion
                description: The availability zones available in the installation's region for use with tenant clusters.
```

PR #172, opened 2019-10-23
Add NOT_SUPPORTED error code

marians:
This adds a new code to be used e. g. if the given operation is not supported on a certain provider.


details/RESPONSE_CODES.md
```diff
@@ -15,3 +15,4 @@ Error messages and standard responses, like after creation of a resource, contai
 - `UNKNOWN_ATTRIBUTE`: Indicates the provided data structure contains unexpected fields.
 - `INVALID_INPUT`: Indicates that the user provided input that is not valid in some way.
 - `UNKNOWN_ERROR`: Indicates something went wrong in unpredictable ways.
+- `NOT_SUPPORTED`: This code is used for an API operation that is implemented in general, but not supported in the current installation for some reason. The reason should be detailed in the `message` part.
```
MarcelMue: Hold on: why is the description so different from the other codes?
Can we use the `Indicates bla` kind of thing?

marians: I tried, but found that it would be bad english.

PR #181, opened 2019-11-06
Add an endpoint for notifying us of clientside exceptions.

oponder:
This goes towards https://github.com/giantswarm/giantswarm/issues/6715.

We want to be notified when an error occurs in Happa (and perhaps gsctl too) and not having to wait until a client tells us about it.


spec/spec.yaml
```diff
@@ -3071,3 +3071,38 @@ paths:
           description: error
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
+
+  /v5/exception-notifications/:
+    post:
+      operationId: createExceptionNotification
+      x-internal: true
+      tags:
+        - exception notification
```
marians: This tag has to be defined in the `tags` section. And it should be plural.
```suggestion
        - exception notifications
```

spec/spec.yaml
```diff
@@ -3071,3 +3071,38 @@ paths:
...
+      summary: Notifies Giant Swarm that an error has occured in one of our clients.
+      description: |
+        This endpoint is used to notify Giant Swarm that an error has occured in one of our clients (Like our Web UI or gsctl). It is not intended to be called manually, our clients are configured to report errors when they occur.
```
marians:
```suggestion
        This endpoint is used to notify Giant Swarm that an error has occured in one of our clients (like our Web UI or gsctl). It is not intended to be called manually, our clients are configured to report errors when they occur.
```

spec/spec.yaml
```diff
@@ -3071,3 +3071,38 @@ paths:
...
+
+        Find us in your slack support channel if you want to contact us about any immediate issues.
```
marians:
```suggestion
        Find us in your Slack support channel if you want to contact us about any immediate issues.
```

spec/spec.yaml
```diff
@@ -3071,3 +3071,38 @@ paths:
...
+        "200":
+          description: Exception notification created
+          schema:
+            $ref: "./definitions.yaml#/definitions/V4GenericResponse"
+          examples:
+            application/json:
+              {
+                "code": "RESOURCE_CREATED",
+                "message": "An exception notification has been sent to Giant Swarm."
+              }
```
marians: Please also add a `default` response description. This is requited to catch all other response codes, like 500.

oponder:
re-ping @marians, fishing for approval from someone. expanding reviewrs to @giantswarm/sig-front-end  and @giantswarm/team-batman

spec/spec.yaml
```diff
@@ -3071,3 +3074,42 @@ paths:
           description: error
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
+
+  /v5/exception-notifications/:
+    post:
+      operationId: createExceptionNotification
```
marians: just noticed: In the operationId we use the verb `add` in cases like `addCluster`. Maybe align this here? (Might be that we are using `create` in some cases, too)

oponder: yeah we have `create` and `add` used without any specific rule. I'll change this to `add` here though while we can.

PR #183, opened 2019-11-06
Set more schema details

marians:
This experimental PR adds some schema specification where details were undefined so far, with the goal to improve automated mocking.

### List of changes

- In `V4ReleaseListItem` set field `active` as required
- in `V4InfoResponse` change type of `workers.count_per_cluster.max` and `workers.count_per_cluster.default` from `number` to `integer`. `number` is not wrong, but `integer` is more precise.
- add `format: byte` to string fields containing base64 data
- add `format: dateTime` to all string field which contain date/time values
- add `format: float` to fields of type `number`
- set `format: semver` to fields holding semantic version numbers
- set `minimum` and `maximum` where possible on integer fields
- set `minItems` and `maxItems` where possible on array fields
- set `minLength` and `maxLength` to string fields where it's actually known 
- set `pattern` to some string fields that allow it

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -347,6 +347,9 @@ definitions:
                   description: |
                     The protocol this port mapping is made for.
                   type: string
+                  oneOf:
+                    - http
+                    - https
 
   # Definition of a v4 cluster node
   V4NodeDefinition:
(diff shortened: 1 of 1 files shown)


oponder:
I'll put this back into the backlog ok! instead of in progress.

marians:
Not pursuing this

PR #185, opened 2019-11-15
Add delete_date attribute

marians:
API spec draft for https://github.com/giantswarm/giantswarm/issues/7516

The idea is to have deleted clusters appear

- in the `GET /v4/clusters/` response
- in the `GET /v4/clusters/{cluster_id}/` response
- in the `GET /v5/clusters/{cluster_id}/` response

as long as they are not fully deleted. The `delete_date` value should represent the date when deletion of the cluster has been requested.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -256,6 +256,10 @@ definitions:
       create_date:
         type: string
         description: Date/time of cluster creation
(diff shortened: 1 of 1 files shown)


xh3b4sd:
Note that only the lister is implemented yet.

PR #191, opened 2019-12-05
Support Azure Multi-AZ

axbarsan:
Epic: https://github.com/giantswarm/giantswarm/labels/Epic%3A%20Multi-AZ

To give a bit of context, we're adding multi-AZ support for Azure, which will not be automatic. The user will be able to pick which AZs he wants to use. While doing that, we also need to keep things compatible with the v4 way of setting availability zones.

My initial solution was creating a custom data structure for multi-AZ, with both a number and a list of zones to use. But after discussing with @oponder, we realized that this solution was not good, because it won't be backwards compatible.

The proposed change is to change the availability zones data type to a string, and for AWS clusters, there would be a single stringified number in there (e.g. `"2"`), but for Azure, we would have a serialized list of the AZs to use (e.g. `"2,4,5`).

As for handling the data differently for AWS and Azure, I suppose this should happen in `clusterclient`, where there would be different requests properties setup for AWS and Azure AZs.

As for supporting legacy requests, with availability zones set as integers, [we can do something like this](https://play.golang.org/p/6ZPsK0VQQvu), which will support both integers and strings.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -183,7 +183,7 @@ definitions:
         description: |
           Number of availability zones a cluster should be spread across. The
           default is provided via the [info](#operation/getInfo) endpoint.
-        type: integer
+        type: string
       scaling:
         type: object
         description: |
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -181,8 +181,9 @@ definitions:
           to use in the new cluster
       availability_zones:
         description: |
-          Number of availability zones a cluster should be spread across. The
-          default is provided via the [info](#operation/getInfo) endpoint.
+          AWS - Stringified number of availability zones a cluster should be spread across.
+          Azure - A serialized list of availability zones a cluster should be spread across
+          The default is provided via the [info](#operation/getInfo) endpoint.
         type: string
       scaling:
         type: object
--- a/spec/spec.yaml
+++ b/spec/spec.yaml
@@ -755,6 +755,10 @@ paths:
         configuration details, like worker node configuration, with node
         specification depending on the provider (e. g. on <span class="badge azure">Azure</span> the
         VM size, or on <span class="badge kvm">KVM</span> the memory size and number of CPU cores).
+        Availability zones are also different between providers (e. g. on
+        <span class="badge azure">Azure</span> the availability zones property
+        is a stringified list of the specific availability zones, but on
+        <span class="badge aws">AWS</span> it will be a stringified number)
 
         One attribute is __mandatory__ upon creation: The `owner`
         attribute must carry the name of the organization the cluster will
@@ -782,7 +786,7 @@ paths:
                 "owner": "myteam",
                 "release_version": "7.1.0",
                 "name": "Example cluster with 3 m5.xlarge nodes",
-                "availability_zones": 2,
+                "availability_zones": "2",
                 "scaling": {"min": 3, "max": 3},
                 "workers": [{"aws": {"instance_type": "m5.xlarge"}}]
               }


spec/definitions.yaml
```diff
@@ -183,7 +183,7 @@ definitions:
         description: |
           Number of availability zones a cluster should be spread across. The
```
fiunchinho: Should this description change? It's just the number of AZ's on AWS, but it's the actual AZ's to use on Azure.

fiunchinho:
I guess we don't have that many options, but having the same parameter to mean different things on different providers doesn't sound ideal. Something to improve in the future I guess.

spec/definitions.yaml
axbarsan: Good catch!

axbarsan:
This is open for discussion, we can find another solution (such as adding another property) together.

oponder:
ping @marians , I'd like to get your opinion on this. 

Mainly, what do you think of switching the type to 'stringified number'? It would be backwards compatible.

Or perhaps it would be best to introduce two new top level objects and deprecate `availability_zones`, like:

`aws.availability_zones`
`azure.availability_zones`

spec/definitions.yaml
```diff
@@ -181,9 +181,10 @@ definitions:
           to use in the new cluster
       availability_zones:
         description: |
-          Number of availability zones a cluster should be spread across. The
-          default is provided via the [info](#operation/getInfo) endpoint.
-        type: integer
+          AWS - Stringified number of availability zones a cluster should be spread across.
+          Azure - A serialized list of availability zones a cluster should be spread across
+          The default is provided via the [info](#operation/getInfo) endpoint.
+        type: string
```
marians: This is a breaking change

marians:
> Mainly, what do you think of switching the type to 'stringified number'? It would be backwards compatible.

Maybe that is true for JavaScript, but definitely not for Go.

I am strongly against introducing a breaking change here and would favour searching for an alternative.

marians:
The suggestion by @oponder to add `azure.availability_zones` works IMO.

Another alternative: Use the existing `workers.azure` object and add a `availability_zones` key to it.

Another thought to consider: In node pools, we offer users to set the number of zones to uses, but let the system select for them. However if they want to select specific zones, they can. Check https://docs.giantswarm.io/api/#operation/addNodePool

In terms of feature parity (let the system behave the same way on all providers, if possible) I would like us to support that on Azure, too. This means...

Specifying the number: 

```json
"availability_zones": {"number": 3}
```

Specifying the exact zones:

```json
"availability_zones": {"zones": ["zone-name1", "zone-name2", "zone-name3"]}
```

axbarsan:
> > Mainly, what do you think of switching the type to 'stringified number'? It would be backwards compatible.
> 
> Maybe that is true for JavaScript, but definitely not for Go.
> 
> I am strongly against introducing a breaking change here and would favour searching for an alternative.

This is possible with an implementation similar to this: 
https://play.golang.org/p/6ZPsK0VQQvu

spec/definitions.yaml
oponder: Depends on where you define the breaking change line. This would be a breaking change for clients that implement the generated go library like gsctl, I suppose, but only when they update that library.

Any existing code or client would still interface correctly, whether it sends a `"3"` or a `3`

So no existing implementations would break. That's what we were thinking of when we thought "not a breaking change".

Anyways just being the devil's advocate here, because I'd prefer not to go for this approach. I'm personally in favour of something like `azure.availability_zones` , because then it also gives us a top level place to add more azure specific things if needed.

oponder:
> Another alternative: Use the existing workers.azure object and add a availability_zones key to it.

Thinking ahead a bit, when we have HA Masters, will they also affected by the availability zones? Or is it always placed in the same zone? Or can we place masters in a different zone than workers?

I think the workers object on create is deprecated anyways right, (it exists in v4, but not in v5) we shouldn't try to keep it around or add more things to it.

axbarsan:
I'm fine with creating a new property just for this, although it will probably be confusing for people trying to use the API, because we'd have 2 different properties with similar names

axbarsan:
I've made another example, with actual converting to the string type:
https://play.golang.org/p/yryXpZyu6wT

oponder:
Yeah I agree, it's pretty confusing, and then there is also the fact that we still have both v4 and v5.

We'd need to talk a bit about the API design here to really sort this out. But we're not willing to invest too much time in API since we want to replace it with direct access to the k8s control plane, and that's where our design efforts should go instead.

I would vote for the new `azure` property and then explaining it well in the docs _and_ validation messages. 

So people creating an azure cluster should get a validation error when they try to use the `availability_zones` key, and it should say please use `azure.availability_zones` instead. 

Currently they could set `availability_zones` to something when creating a cluster and it would just not have any effect right? So that wouldn't be a breaking change, since I doubt anyone was filling anything in for `availability_zones` on azure.

Hopefully we'll see `api` get dropped entirely sooner rather than later.

axbarsan:
I'm ok with either creating a new property of using the hackish string/int way 👍

marians:
> Thinking ahead a bit, when we have HA Masters, will they also affected by the availability zones? Or is it always placed in the same zone? Or can we place masters in a different zone than workers?

HA Masters AZ assignment is independent of worker nodes. It will exist only in v5 btw.

> I think the workers object on create is deprecated anyways right, (it exists in v4, but not in v5) we shouldn't try to keep it around or add more things to it.

That is true for all of v4. It will be deprecated as soon as Azure supports node pools.

> So people creating an azure cluster should get a validation error when they try to use the availability_zones key, and it should say please use azure.availability_zones instead.
> Currently they could set availability_zones to something when creating a cluster and it would just not have any effect right? So that wouldn't be a breaking change, since I doubt anyone was filling anything in for availability_zones on azure.

Yeah, in fact that is the case. We don't explain it in the API docs but `availability_zones` should not have any effect on Azure currently, so keeping it that way is not a breaking change.

oponder:
Alright, let's roll with the `azure.availability_zones` then?

axbarsan:
So as @marians suggested in the UX meeting today, the goal is to keep a consistent implementation across our providers, and to keep the API docs as easy to understand as possible.

To achieve this, we'll have both Azure and AWS V4 use the same number `availability_zones` property. Which means the user will not be able to select their own preferred Availability Zones for Azure (as it is with AWS AZs right now).

As `azure-operator` has to receive a list of the availability zones to use, we will assign them randomly in the API.

PR #195, opened 2020-01-17
Add ServiceAccount, Role, RoleBinding, and PSP for using EmptyDir

marians:
Towards https://github.com/giantswarm/giantswarm/issues/7349

**This PR does not contain any changes regarding the Giant Swarm API specification. The focus is on the technicalities of publishing the documentation.**

This PR should enable the app to use an EmptyDir volume for the nginx cache. It's derived from the working docs deployment in gollum/35ac3.

### DIFF
--- a/helm/api-spec-app/templates/psp-rbac.yaml
+++ b/helm/api-spec-app/templates/psp-rbac.yaml
@@ -1,0 +1,60 @@
+---
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+    name: {{ .Values.name }}
+spec:
+    allowPrivilegeEscalation: false
+    fsGroup:
+    ranges:
+    - max: 65535
+        min: 1
+    rule: MustRunAs
+    runAsUser:
+    rule: MustRunAsNonRoot
+    seLinux:
+    rule: RunAsAny
+    supplementalGroups:
+    ranges:
+    - max: 65535
+        min: 1
+    rule: MustRunAs
+    volumes:
+    - secret
+    - configMap
+    - emptyDir
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+    namespace: {{ .Release.Namespace }}
+    name: {{ .Values.name }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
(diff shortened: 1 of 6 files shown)


marians:
Would be kind if you could just check for any typos

helm/api-spec-app/templates/psp-rbac.yaml
```diff
@@ -0,0 +1,60 @@
+---
+apiVersion: extensions/v1beta1
```
oponder:
```suggestion
apiVersion: policy/v1beta1
```

helm/api-spec-app/templates/psp-rbac.yaml
```diff
@@ -0,0 +1,60 @@
...
+kind: PodSecurityPolicy
+metadata:
+    name: {{ .Values.name }}
+spec:
+    allowPrivilegeEscalation: false
+    fsGroup:
```
oponder: I feel like theres some indentation wrong here.

oponder: and in other places (like seLinux for example)

helm/api-spec-app/templates/psp-rbac.yaml
```diff
@@ -0,0 +1,60 @@
...
+    rule: RunAsAny
+    supplementalGroups:
+    ranges:
+    - max: 65535
+        min: 1
+    rule: MustRunAs
+    volumes:
+    - secret
+    - configMap
+    - emptyDir
```
calvix: the indentation in psp is broken, check this for reference https://github.com/giantswarm/happa/blob/master/helm/happa-chart/templates/psp.yaml

marians: Fixed

helm/api-spec-app/templates/psp-rbac.yaml
marians: Done

helm/api-spec-app/templates/psp-rbac.yaml
marians: Done

marians:
@oponder Can you re-think your review, please? 🌻

PR #196, opened 2020-01-28
add /v5/clusters/{cluster_id}/labels/ api-spec

ubergesundheit:
Towards giantswarm/giantswarm#1693, Related: giantswarm/api#979

This pull request adds documentation about endpoints to list and set labels for a v5 cluster.


spec/spec.yaml
```diff
@@ -77,6 +77,11 @@ tags:
       A release provides _components_, like for example Kubernetes. For each
       release the contained components are listed. Changes in components are
       detailed in the _changelog_ of a release.
+  - name: cluster labels
+    description: |
+      Cluster labels allow users to group and tag clusters with own key value pairs.
+
+      Note: Cluster labeling is currently (as of February 2020) only available on <span class="badge aws">AWS</span> to selected Giant Swarm customers.
```
MarcelMue: Why this note?

ubergesundheit: Following the discussion in giantswarm/api#979 and [this comment](https://github.com/giantswarm/giantswarm/issues/1693#issuecomment-578764717) I decided to add this for now only for v5 clusters

MarcelMue: I understand - I would be happy if we at least bring it to other providers as well.

MarcelMue: But okay to sort AWS first.

ubergesundheit: Yeah but now as you wrote:
> But ok to sort AWS first
I think this notice should be rewritten to something like

`Cluster labeling is currently (as of February 2020) only available on cluster releases 10 or higher on <span class="badge aws">AWS</span>.`

spec/spec.yaml
```diff
@@ -77,6 +77,11 @@ tags:
       A release provides _components_, like for example Kubernetes. For each
       release the contained components are listed. Changes in components are
       detailed in the _changelog_ of a release.
+  - name: cluster labels
+    description: |
+      Cluster labels allow users to group and tag clusters with own key value pairs.
```
marians:
```suggestion
      Cluster labels allow users to group and tag clusters with custom key-value-pairs.
```

spec/spec.yaml
```diff
@@ -77,6 +77,11 @@ tags:
...
+
+      Note: Cluster labeling is currently (as of February 2020) only available on cluster releases 10 or higher on <span class="badge aws">AWS</span>.
```
marians:
```suggestion
      Note: Cluster labelling is currently (as of February 2020) only available on cluster releases 10 or higher on <span class="badge aws">AWS</span>.
```

marians:
Question: Did you consider delivering the labels with the base cluster information (`GET /v5/clusters/:clusterid/`)? Why do we provide a specific endpoint for the labels?

ubergesundheit:
Labels could be returned with the cluster information request as well! 👍 I’ll make the changes next week

ubergesundheit:
Following the question of @marians I've added cluster labels to the cluster details request for v5. It now contains a `labels` field. Should this also appear for v4 clusters but just always contain `null` for consistency?

marians:
> Should this also appear for v4 clusters but just always contain null for consistency?

I think adding a new attribute to the v4 endpoint that will never hold any data isn't needed. I can't think of any benefit.

spec/definitions.yaml
```diff
@@ -1137,3 +1141,32 @@ definitions:
...
+        $ref: "./definitions.yaml#/definitions/V5ClusterLabels"
+
+  # cluster labels
+  V5ClusterLabels:
+    type: object
+    title: Labels object
+    description: Object containing keys with string values representing the labels attached to the cluster
+    additionalProperties:
+      title: Labels
+      description: Key value pairs representing the labels attached to the cluster
```
MarcelMue, line 1161: Super nit pick: Would be cool to have some order consistency - does the `description:` field come first / last / second :P

ubergesundheit:
This PR now includes the following additions/changes:

- `/v5/clusters/{cluster_id}/labels/` for reading/changing cluster labels
- `/v4/clusters/` will include labels for v5 clusters
- `/v5/clusters/{cluster_id}/` will include labels
- `/v5/clusters/by_labels` with method POST can be used to list v5 clusters based on labels


Once this is approved I'll start working on actual api implementation

spec/definitions.yaml
ubergesundheit: Changed it to `type`, `title`, `description`

ubergesundheit:
@giantswarm/sig-ux can I get some :eyes: and :heavy_check_mark: ?

spec/spec.yaml
```diff
@@ -77,6 +77,11 @@ tags:
       A release provides _components_, like for example Kubernetes. For each
       release the contained components are listed. Changes in components are
       detailed in the _changelog_ of a release.
+  - name: cluster labels
+    description: |
+      Cluster labels allow users to group and tag clusters with custom key-value-pairs.
+
+      Note: Cluster labelling is currently (as of February 2020) only available on cluster releases 10 or higher on <span class="badge aws">AWS</span>.
```
oponder, line 84:
```suggestion
      Note: Cluster labelling is currently only available on release version 10 or higher on <span class="badge aws">AWS</span>.
```

We can update this with corrected info when the time comes, I don't think we need to have the date in here necessarily

spec/spec.yaml
```diff
@@ -3936,6 +3952,180 @@ paths:
...
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
 
+  /v5/clusters/{cluster_id}/labels/:
+    get:
+      operationId: getClusterLabels
+      tags:
+        - cluster labels
+      summary: Get cluster labels
+      description: |
+        Return labels attached to the supplied cluster.
```
oponder, line 3962:
```suggestion
        Returns the labels that this cluster is labelled with
```

`attached` is kind of a new domain word, I don't think we need to introduce that. When we want to add a label to a cluster, we set the labels (since thats what we're calling it in the put action), and the cluster is labelled.

spec/spec.yaml
```diff
@@ -3936,6 +3952,180 @@ paths:
...
+      parameters:
+        - $ref: "./parameters.yaml#/parameters/ClusterIdPathParameter"
+      responses:
+        "200":
+          description: Labels attached to this cluster
```
oponder, line 3967:
```suggestion
          description: Cluster labels
```

This shows up in a pretty small place in the docs ui, its describing the response.

spec/spec.yaml
```diff
@@ -3936,6 +3952,180 @@ paths:
...
+    post:
+      operationId: getV5ClustersByLabel
+      tags:
+        - clusters
+      summary: Get clusters by labels (v5)
+      description: |
+        This operation fetches a list of node pool clusters based
+        on a label selector.
+
+        The operation accepts label selectors analogue to `kubectl get -l`
```
oponder, line 4058:
```suggestion
        The operation accepts label selectors in the same way that `kubectl get -l` does
```

I try to make language in docs as simple as possible, even if it means a few more words, I think it ends up faster for most people to parse.

spec/spec.yaml
```diff
@@ -3936,6 +3952,180 @@ paths:
...
+        ([kubernetes label selectors description](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors))
+        for listing clusters based on their labels.
+
+        The result depends on the permissions of the user.
+        A normal user can search over all the clusters the user has access
+        to, via organization membership.
```
oponder, line 4064:
```suggestion
        A normal user can search over all the clusters that they have access
        to, based on their organization memberships.
```

spec/spec.yaml
```diff
@@ -3936,6 +3952,180 @@ paths:
...
+        A user with admin permission searches over a list of all existing
+        clusters.
```
oponder, line 4066:
```suggestion
        Admin users however, will search over all existing clusters.
```

spec/spec.yaml
```diff
@@ -3936,6 +3952,180 @@ paths:
...
+
+        The result array items are sparse representations of the cluster
+        objects. To fetch more details on a cluster, use the
+        [getClusterV5](#operation/getClusterV5) operation.
```
oponder, line 4070:
```suggestion
        The resulting array contains a sparse representation of the cluster objects. To fetch more details on a cluster, use 
        the [getClusterV5](#operation/getClusterV5) operation.
```

spec/spec.yaml
```diff
@@ -77,6 +77,11 @@ tags:
       A release provides _components_, like for example Kubernetes. For each
       release the contained components are listed. Changes in components are
       detailed in the _changelog_ of a release.
+  - name: cluster labels
```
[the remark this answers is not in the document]
ubergesundheit: yeah makes sense!

PR #199, opened 2020-03-25
Add support for spot instances

MarcelMue:

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -945,7 +945,7 @@ definitions:
             type: string
             description: |
               EC2 instance type to use for all nodes in the node pool. _(Validated against available instance types.)_
-          use_alike:
+          use_alike_instance_types:
             type: boolean
             description: |
               If true, instances alike the instance_type will be used.
@@ -976,7 +976,7 @@ definitions:
             type: string
             description: |
               EC2 instance type used by all nodes in this pool
-          use_alike:
+          use_alike_instance_types:
             type: boolean
             description: |
               If true, instances alike the instance_type will be used.
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -932,6 +932,10 @@ definitions:
             type: string
             description: |
               EC2 instance type to use for all nodes in the node pool. _(Validated against available instance types.)_
+          spot_instance_enabled:
+            type: boolean
+            description: |
+              If true, the node pool will be filled with spot instances.
 
   # Specification of a node in a node pool after creation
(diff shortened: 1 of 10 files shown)


marians:
I would prefer plural `spot_instances_enabled` over singular `spot_instance_enabled`, since a node pool is normally about a group of instances.

spec/definitions.yaml
```diff
@@ -928,14 +928,27 @@ definitions:
...
           Attributes specific to the AWS provider
         properties:
+          instance_distribution:
+            description: |
+              Attributes defining the instance distribution in a node pool.
+            type: object
+            properties:
+              on_demand_base_capacity:
+                description: |
+                  Base capacity of on demand machines.
```
marians: The description should explain what exactly the number would mean. Is it a percentage or a number of machines?

spec/definitions.yaml
```diff
@@ -928,14 +928,27 @@ definitions:
...
+              on_demand_percentage_above_base_capacity:
+                description: |
+                  Percentage of on demand instances above the base capacity.
+                type: integer
           instance_type:
             type: string
             description: |
               EC2 instance type to use for all nodes in the node pool. _(Validated against available instance types.)_
-          spot_instance_enabled:
+          use_alike:
```
marians: I would like this to be more specific, e. g. `use_alike_instance_types`.

marians:
Actually, I don't feel as strongly after some more thinking. However, the comment

```
# Specification of a node in a node pool to be created
V5AddNodePoolRequestNodeSpec:
```

should be adapted to reflect the actual purpose better.

marians:
Related to https://github.com/giantswarm/giantswarm/issues/8347

spec/definitions.yaml
```diff
@@ -1000,12 +1037,21 @@ definitions:
         description: Information on the current size and status of the node pool
         type: object
         properties:
+          instance_types:
```
MarcelMue: @marians I couldn't think of a good structure here which doesn't totally change this stuff. Do you have an idea?

marians:
Consistency thing: property names should use camel case. E. g. `instance_distribution` should be `instanceDistribution`.

spec/definitions.yaml
marians: This mention went past me, sorry. Don't see the problem here actually.

marians:
FYI: With the latest commits I also added `default` attributes to inform users about what we set as a default value if the attribute is not specified when creating a node pool.

PR #200, opened 2020-04-09
Add spec and info for the 503 response in the addKeyPair operation

marians:
Towards https://github.com/giantswarm/giantswarm/issues/9856

This adds spec and info for the 503 response in the addKeyPair operation.


marians:
All right, this is waiting for implementation in the `api`.

PR #201, opened 2020-04-30
Add 400 error for invalid instance type

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/3046

This adds a 400 status code error in case the provided instance type while creating a cluster is not supported.

### DIFF
--- a/spec/spec.yaml
+++ b/spec/spec.yaml
@@ -815,7 +815,7 @@ paths:
                 "message": "A new cluster has been created with ID 'wqtlq'"
               }
         "400":
-          description: Instance type not found
+          description: Invalid request
           schema:
             $ref: "./definitions.yaml#/definitions/V4GenericResponse"
           examples:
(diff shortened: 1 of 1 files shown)


spec/spec.yaml
```diff
@@ -814,6 +814,16 @@ paths:
                 "code": "RESOURCE_CREATED",
                 "message": "A new cluster has been created with ID 'wqtlq'"
               }
+        "400":
+          description: Instance type not found
```
marians, line 818:
```suggestion
          description: Invalid request
```

Since the 400 response code is used for all sorts of invalid inputs, the headline should reflect that.

spec/spec.yaml
```diff
@@ -3523,6 +3533,16 @@ paths:
                 },
                 "nodepools": []
               }
+        "400":
+          description: Instance type not found
```
marians, line 3537:
```suggestion
          description: Invalid request
```

See above

PR #202, opened 2020-05-04
Update spec for master nodes high availability

marians:
For [HA Masters](https://github.com/giantswarm/roadmap/issues/58). Closes https://github.com/giantswarm/giantswarm/issues/7901

## Notable changes

- For v5 cluster creation, the `master` property gets deprecated, to be replaced by the new `master_nodes` property. There will be a deprecation phase where the old form can still be used to create clusters with single master nodes.
- For getting v5 cluster details, likewise the `master` property gets deprecated, eventually to be replaced by the new `master_nodes` property. During the deprecation phase, both properties will be served in case the cluster has only one master node.
- The `modifyClusterV5` request body is changed to support the switch from single to HA masters.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -1202,6 +1202,7 @@ definitions:
             type: array
             description: |
               Availability zones of the master node(s) of this cluster.
+            x-nullable: true
             items:
               type: string
           num_ready:
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -105,6 +105,13 @@ definitions:
               release_version_minimum:
                 type: string
                 description: The minimum release version number required to have support for node pools.
+          spot_instances:
+            type: object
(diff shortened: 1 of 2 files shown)


marians:
> We won't allow them settings the AZs on the API, is that intended? Guess it's due to the deprecation but feels strange to remove some functionality in one layer

We are not removing that functionality right away. And until the `master` property is being removed, we can decide if and how to transfer the AZ placement functionality to the new schema.

marians:
**Change notice:** Added the attribute `master_nodes.num_ready` in the cluster details response. This is a direct mapping of the G8sControlPlane CR attribute `.status.readyReplicas`.

marians:
**Change notice:** Added the `high_availability` attribute to the `modifyClusterV5` request, to allow API users to switch a cluster from single to multiple master nodes.

PR #208, opened 2020-05-12
Change scaling example to state 'invalid input'

axbarsan:
This is a more suitable code.


axbarsan:
Merging without approval

PR #210, opened 2020-06-04
Align ConfigMap case, explain user_config editing.

oponder:
A customer got a bit confused about the `user_config` not being editable, and thinking that meant that config in general was not editable. I tried to explain that better.

I also aligned the capitalization of ConfigMap throughout the descriptions in the spec.


oponder:
See: https://gigantic.slack.com/archives/CE92C4BST/p1590487458193700

PR #211, opened 2020-06-12
Return release state directly

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/11544

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -105,6 +105,13 @@ definitions:
               release_version_minimum:
                 type: string
                 description: The minimum release version number required to have support for node pools.
+          spot_instances:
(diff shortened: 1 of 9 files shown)


spec/definitions.yaml
```diff
@@ -715,6 +715,13 @@ definitions:
             version:
               type: string
               description: Version number of the component
+      state:
+        type: string
```
marians, line 719:
```suggestion
        type: string
        enum: ["active", "wip", "deprecated"]
```

marians: Please also add the `state` attribute to any response examples where it applies.

tfussell:
I don't have time to work on this.

PR #212, opened 2020-07-09
Add Azure specific fields to V5GetNodePoolResponseNodeSpec

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/9.

What's in the box:
- Adding Azure-specific fields to the API response when listing node pools

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -999,6 +999,7 @@ definitions:
               Azure virtual machine size used by all nodes in this pool.
       aws:
         type: object
+        x-nullable: true
         description: |
           Attributes specific to the AWS provider
         properties:
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -989,6 +989,7 @@ definitions:
     properties:
       azure:
         type: object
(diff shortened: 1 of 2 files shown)


spec/definitions.yaml
```diff
@@ -987,6 +987,15 @@ definitions:
   V5GetNodePoolResponseNodeSpec:
     type: object
     properties:
+      azure:
+        type: object
```
axbarsan, line 991:
```suggestion
        type: object
        x-nullable: true
```

spec/definitions.yaml
```diff
@@ -987,6 +987,15 @@ definitions:
...
+        description: |
+          Attributes specific to the Azure provider
+        properties:
+          vm_size:
+            type: string
+            description: |
+              Azure virtual machine size used by all nodes in this pool.
       aws:
         type: object
```
axbarsan, line 1000:
```suggestion
        type: object
        x-nullable: true
```

nprokopic:
> I believe these can be `null`, right?

You are correct!

PR #213, opened 2020-07-15
Add Azure spec to node pool creation

tuommaki:

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -984,6 +984,7 @@ definitions:
               Added with AWS release v11.2.0.
       azure:
         type: object
+        x-nullable: true
         description: |
           Attributes specific to the Azure provider
         properties:
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -942,6 +942,7 @@ definitions:
     properties:
       aws:
         type: object
+        x-nullable: true
         description: |
           Attributes specific to the AWS provider
         properties:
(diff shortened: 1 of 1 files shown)


spec/definitions.yaml
```diff
@@ -982,6 +982,15 @@ definitions:
               increase the likelihood to get instances for this pool, especially spot instances at
               a low rate. If false, only the exact type set as `instance_type` is used.
               Added with AWS release v11.2.0.
+      azure:
+        type: object
```
axbarsan, line 986: This can be `null`, right? If correct, then we should add it to the `aws` field, as well

```suggestion
        type: object
        x-nullable: true
```

tuommaki: Oh, yes. Thanks! Good catch!

spec/definitions.yaml
```diff
@@ -942,6 +942,7 @@ definitions:
     properties:
       aws:
         type: object
+        x-nullable: true
```
tuommaki, line 945: @marians Could this introduce problems for existing clients/customers? On AWS I'd expect the value to always be there and on Azure there are no existing clients yet. Can I just merge this or do I need to do any additional work here?

marians: Changing a request body spec in this way makes the API more tolerant and should not cause any problems to clients.

marians: Merging should only happen after the change specified here is implemented and the API is rolled out, so that the api-spec `master` always represents what users can do with the API.

tuommaki: Ack. Thanks! :+1:

PR #216, opened 2020-08-05
Make release status required

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12567

This PR makes the release status required, so it won't be annotated as `omitempty` in the generated `gsclientgen` data structure.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -681,6 +681,7 @@ definitions:
         description: Date and time of the release creation
       active:
         type: boolean
+        x-nullable: false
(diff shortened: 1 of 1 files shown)


spec/definitions.yaml
```diff
@@ -681,6 +681,7 @@ definitions:
         description: Date and time of the release creation
       active:
         type: boolean
+        x-nullable: false
```
axbarsan, line 684: This prevents the `active` value from being a pointer in the generate Go code

PR #217, opened 2020-08-07
document cluster creation possible with deprecated versions

ubergesundheit:
Towards giantswarm/giantswarm#12489

This changes the description text of the `active` field of release response items to make it clear that cluster creation is possible but not recommended.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -683,9 +683,10 @@ definitions:
         type: boolean
         x-nullable: false
         description: |
-          If true, the version is available for new clusters and cluster
+          If `true`, the version is recommended for use in cluster creation and cluster
           upgrades. Older versions become unavailable and thus have the
-          value `false` here.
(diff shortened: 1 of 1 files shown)


spec/definitions.yaml
```diff
@@ -683,9 +683,10 @@ definitions:
         type: boolean
         x-nullable: false
         description: |
-          If true, the version is available for new clusters and cluster
+          If `true`, the version is recommended for use in cluster creation and cluster
           upgrades. Older versions become unavailable and thus have the
-          value `false` here.
+          value `false` here. Cluster creation with this version is still possible
```
marians, line 688: "still possible", but under the condition that there is at least one cluster running with the given release version in the installation.

ubergesundheit: That’s not how it’s currently implemented. Cluster creation is possible with every release in the CP

marians (approved):

PR #221, opened 2020-10-01
Add property for supported K8s versions in the info response

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/13439

Add new `supported_k8s_version_skews` property in the info response, for displaying the supported K8s versions. Versions older than the ones specified are EOL and should not be used for new clusters.

This PR will be merged once the API implementation is done and released.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -76,11 +76,22 @@ definitions:
           datacenter:
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
-          supported_k8s_version_skews:
-            description: The supported Kubernetes version skews.
+          kubernetes_versions:
+            description: A list of the available Kubernetes versions available, and their EOL dates.
             type: array
             items:
-              type: string
+              type: object
+              required:
+                - minor_version
+                - eol_date
+              properties:
+                minor_version:
+                  description: The version of the Kubernetes release.
+                  type: string
+                eol_date:
+                  description: The date when the release becomes EOL.
+                  type: string
+                  format: date-time
           availability_zones:
             type: object
             description: Number of availability zones which a cluster can be spread across.
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -77,7 +77,7 @@ definitions:
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
           kubernetes_versions:
-            description: A list of the available Kubernetes versions available, and their EOL dates.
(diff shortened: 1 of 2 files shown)


marians:
Thanks for taking this on! Some feedback:

The name `supported_k8s_version_skews` is a bit much, IMO. And it is very specific, which makes it not very extensible. 

How about calling the top level attribute `kubernetes_versions`?

The value could be an array containing a version number and an EOL date. The date would also give us the ability to warn beforehand, if we have the EOL date before the fact.

```json
"kubernetes_versions": [
   {
      "minor_version": "1.16",
      "eol_date": "2020-01-01"
   }
]
```

This structure would be more extensible to store additional kubernetes version information.

axbarsan:
> How about calling the top level attribute `kubernetes_versions`?
> 
> The value could be an array containing a version number and an EOL date. The date would also give us the ability to warn beforehand, if we have the EOL date before the fact.
> 
> ```json
> "kubernetes_versions": [
>    {
>       "minor_version": "1.16",
>       "eol_date": "2020-01-01"
>    }
> ]
> ```
> 
> This structure would be more extensible to store additional kubernetes version information.

I really like that 👍🏻 Thank you for the review

spec/definitions.yaml
```diff
@@ -76,11 +76,22 @@ definitions:
           datacenter:
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
-          supported_k8s_version_skews:
-            description: The supported Kubernetes version skews.
+          kubernetes_versions:
+            description: A list of the available Kubernetes versions available, and their EOL dates.
```
marians, line 80:
```suggestion
            description: Information on some kubernetes versions and their end of life dates.
```

spec/spec.yaml
```diff
@@ -234,8 +234,15 @@ paths:
...
-              "1.17", "1.18", "1.19"
+            "kubernetes_versions": [
+              {
+                "minor_version": "1.15",
+                "eol_date": "2020-03-24T00:00:47.215Z"
+              },
+              {
+                "minor_version": "1.16",
+                "eol_date": "2020-08-26T00:00:47.215Z"
+              }
```
marians, line 245: I would prefer to not have the time part included (`T00:00:47.215Z`). We don't have that information, and a date is accurate enough.

axbarsan: I changed the format to `date`, so we won't have to include the time 👍🏻

marians (approved):

pipo02mix (approved):

PR #222, opened 2020-11-16
Allow set min nodepool to 0

paurosello:

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -948,6 +948,7 @@ definitions:
           min:
             description: Minimum number of nodes in the pool.
             type: integer
+            x-nullable: true
           max:
             description: Maximum number of nodes in the pool.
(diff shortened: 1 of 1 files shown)


paurosello:
Towards https://github.com/giantswarm/roadmap/issues/234

calvix (approved):

xh3b4sd (approved):

pipo02mix (approved):

marians:
Do we need `x-nullable` on the attribute to be able to accept an int value of zero?

paurosello:
> Do we need `x-nullable` on the attribute to be able to accept an int value of zero?

Yup, not really intuitive but based on the research with Tim and the tests we did this is the only way

marians:
Fine. Doesn't hurt from the API users perspective.

PR #223, opened 2020-11-19
Added new fields to the v5/addCluster call to better specify how avai…

whites11:
Towards https://github.com/giantswarm/giantswarm/issues/13842

When I am creating a CAPI cluster on Azure I would like to be able to select between 3 modes for assigning availability zones to the master nodes:

- automated: system assigns random availability zones to each master node
- manual: users select availability zones to be used for masters
- none: no availability zones have to be used.

This PR adds the needed, backward compatible fields to make this feature possible.

### DIFF
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -1192,7 +1192,7 @@ definitions:
               This field allows manually specifying the availability zone(s) where the master node(s)
               should be created in.
 
-              This parameter is supported on Azure clusters only.
+              This parameter is currently only supported on <span class="badge azure">Azure</span>.
           azure:
             type: object
             description: |
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -76,6 +76,22 @@ definitions:
           datacenter:
             description: Identifier of the datacenter or cloud provider region, e. g. "eu-west-1"
             type: string
+          kubernetes_versions:
+            description: Information on some kubernetes versions and their end of life dates.
+            type: array
+            items:
+              type: object
+              required:
+                - minor_version
+                - eol_date
+              properties:
+                minor_version:
+                  description: The version of the Kubernetes release.
+                  type: string
+                eol_date:
+                  description: The date when the release becomes EOL.
+                  type: string
+                  format: date
           availability_zones:
             type: object
             description: Number of availability zones which a cluster can be spread across.
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -1183,6 +1183,29 @@ definitions:
 
               When `false`, only one master node will be created, also in a randomly
               selected availability zone.
+          availability_zones_selection_mode:
+            type: string
+            default: auto
+            description: |
+              Specifies how the availability zone(s) for the master node(s) should be selected.
+              Possible values are `auto`, `manual` and `disabled`
+
+              When set to `auto` the system will randomly assign an availability zone to each
+              master node.
+
+              When set to `manual` the request needs to contain an additional field `availability_zones`
+              containing the list of desired availability zones the master nodes should be placed in.
+
+              When set to `disabled` the master nodes will not be placed in any availability zone
+              (supported on Azure clusters only).
+          availability_zones:
+            type: array
+            items:
+              type: string
+            description: |
+              When `availability_zones_selection_mode` is set to `manual`, this field becomes
+              mandatory and allow manually specifying the availability zone(s) where the master node(s)
+              should be created in.
   V5ClusterDetailsResponse:
     type: object
     properties:
--- a/spec/definitions.yaml
+++ b/spec/definitions.yaml
@@ -1196,7 +1196,7 @@ definitions:
           azure:
             type: object
             description: |
-              Attributes specific to nodes running on Microsoft Azure
+              Attributes specific to nodes running on <span class="badge azure">Azure</span>.
             properties:
               availability_zones_unspecified:
                 type: boolean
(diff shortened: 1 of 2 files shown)


spec/definitions.yaml
```diff
@@ -1183,6 +1183,29 @@ definitions:
...
+              When set to `disabled` the master nodes will not be placed in any availability zone
+              (supported on Azure clusters only).
+          availability_zones:
+            type: array
+            items:
+              type: string
+            description: |
+              When `availability_zones_selection_mode` is set to `manual`, this field becomes
+              mandatory and allow manually specifying the availability zone(s) where the master node(s)
+              should be created in.
```
marians, line 1208: Were we should also mention that this is only supported on Azure

whites11: Is it? I mean, this can be supported in any cloud provider.

marians: Can, but it isn't the case. On AWS users cannot specify the master AZs currently.

marians:
I still want to give the naming of the attributes some thought.

Have you thought about adding the same attributes to the cluster details response? (https://docs.giantswarm.io/api/#operation/getClusterV5)

spec/definitions.yaml
whites11: ok added note

whites11:
> I still want to give the naming of the attributes some thought.
> 
> Have you thought about adding the same attributes to the cluster details response? ([docs.giantswarm.io/api/#operation/getClusterV5](https://docs.giantswarm.io/api/#operation/getClusterV5))

Yes, I thought about that and concluded they are not necessary.
After the cluster is created, the user is interested in the actual availability zones that the master node(s) are running in. I reckon there is zero interest in understanding how those availability zones have been chosen at cluster creation time.

WDYT?

marians:
I am thinking that we don't need a specific flag and value to express the _automatic assignment by Giant Swarm logic_ case (which is the default). Clients can simply omit the .master_nodes.availability_zones array, set it to `null` (if ), or leave it empty `[]`. 

Also, for manual assignment, we can simply accept something like this:

```json
{
  "master_nodes": {
    "availability_zones": ["a", "b", "c"]
  }
}
```

So we only need a way to express that we want no AZ assignment at all. And this is Azure specific. For provider specific attribute, e. g. in https://docs.giantswarm.io/api/#operation/addNodePool we have created sub-objects named after the provider.

Here this could look like this:

```json
{
  "master_nodes": {
    "azure": {
        "availability_zone_unspecified": true
    }
  }
}
```

We have to document that when `.master_nodes.azure.availability_zone_unspecified` is `true`, then either (A) `.master_nodes.availability_zones` must be empty (otherwise an error 400 will occur) or (B) any `.master_nodes.availability_zones` value will be ignored.

tuommaki:
I'm not blocking here, but I feel that `.master_nodes.azure.availability_zones_unspecified` flag is unnecessary and repetitive. In JSON interfaces I treat `null` the same way as in [SQL](https://en.wikipedia.org/wiki/Null_(SQL)) because memory pointer doesn't make sense in that context. Hence it effectively means `not specified` / `missing` value.

Semantically I'd say that all Azure states can be expressed with `availability_zones` list given that it's nullable:
- **AZs specified**: `availability_zones: ["1", "3"]`
- **AZs unspecified (auto-generate)**: `availability_zones: null`
- **No AZs specified explicitly**: `availability_zones: []`

WDYT?

whites11:
> I'm not blocking here, but I feel that `.master_nodes.azure.availability_zones_unspecified` flag is unnecessary and repetitive. In JSON interfaces I treat `null` the same way as in [SQL](https://en.wikipedia.org/wiki/Null_(SQL)) because memory pointer doesn't make sense in that context. Hence it effectively means `not specified` / `missing` value.
> 
> Semantically I'd say that all Azure states can be expressed with `availability_zones` list given that it's nullable:
> 
> * **AZs specified**: `availability_zones: ["1", "3"]`
> * **AZs unspecified (auto-generate)**: `availability_zones: null`
> * **No AZs specified explicitly**: `availability_zones: []`
> 
> WDYT?

I like it a lot! @marians ?

marians:
I think the difference between `availability_zones: null` and `availability_zones: []` is very subtle and I can imagine API users not noticing the difference, because it is against intuition.

Any thoughts from somebody else in @giantswarm/sig-ux ?

oponder:
I think Marian's suggestion is more explicit and self documenting.

@tuommaki's suggestion is more economical and compact, but from a ux point of view, I find it's a bit easier to explain and reason about Marian's proposal, as well as easier to not accidentally make a mistake or forget which is which.

whites11:
> I think Marian's suggestion is more explicit and self documenting.
> 
> @tuommaki's suggestion is more economical and compact, but from a ux point of view, I find it's a bit easier to explain and reason about Marian's proposal, as well as easier to not accidentally make a mistake or forget which is which.

OK thanks for the feedback I will update the PR with Marian's proposal

spec/definitions.yaml
```diff
@@ -1184,6 +1184,25 @@ definitions:
...
               selected availability zone.
+          availability_zones:
+            type: array
+            items:
+              type: string
+            description: |
+              This field allows manually specifying the availability zone(s) where the master node(s)
+              should be created in.
+
+              This parameter is supported on Azure clusters only.
```
marians, line 1195:
```suggestion
              This parameter is currently only supported on Azure.
```

spec/definitions.yaml
```diff
@@ -1184,6 +1184,25 @@ definitions:
...
+          azure:
+            type: object
+            description: |
+              Attributes specific to nodes running on Microsoft Azure
```
marians, line 1199:
```suggestion
              Attributes specific to nodes running on Azure.
```

marians (approved):
Tried to beautify the descriptions a bit. LGTM!

PR #227, opened 2021-01-13
Remove the unnecessary mention of the V5 API being supported by AWS only

axbarsan:
This mention is no longer the case, and unnecessary, in my opinion. What do you think?

### DIFF
--- a/spec/spec.yaml
+++ b/spec/spec.yaml
@@ -59,9 +59,6 @@ tags:
       description: "User guide: Accessing Pods and Services from the Outside"
   - name: node pools
     description: |
-      This concept is introduced with v5 of this API and is currently (as of October 2019) only
-      available on <span class="badge aws">AWS</span> to selected Giant Swarm customers.
-
       Node pools are groups of worker nodes. All worker nodes in a
       node pool share the same configuration. Also scaling is controlled on a pool
       level. This allows to fulfill a variety of different workload requirements


whites11:
Maybe better to a specify KVM is not supported or that AWS and azure are supported?

axbarsan:
But since platform support is so nicely displayed in the docs, does it make sense to copy-paste that info into the api spec, as well?

https://docs.giantswarm.io/basics/nodepools/

whites11:
question: what happens if you call the API endpoint for KVM cluster? Does it fail in a clear way? Such as "node pools don't work on KVM"?

tuommaki:
> question: what happens if you call the API endpoint for KVM cluster? Does it fail in a clear way? Such as "node pools don't work on KVM"?

Depends. Read/List operations most probably return 404 while other operations are clearer on that. E.g. cluster creation here:
https://github.com/giantswarm/api/blob/master/service/cluster/creator/service.go#L207

tuommaki:
> does it make sense to copy-paste that info into the api spec, as well?

I don't think so. `api-spec` specifies the API available. Documentation specifies which provider supports which features. Bad analogy: You buy a mechanic's tool (e.g. torque wrench) which is accompanied with manual how to use it. It doesn't describe products you can use it on (nails?).

tuommaki (approved):

PR #231, opened 2021-02-10
Set delete dates to strings

oponder:
The generated client, is somehow unable to convert the dates it is getting. 

Gotta find out what the consequence of this will be for gsctl.

towards: https://github.com/giantswarm/giantswarm/issues/15315


oponder:
FYI @axbarsan I started on this, but won't try to pull it all the way through tonight.

axbarsan (approved):

PR #243, opened 2024-02-19
Align files

architectbot:
This PR is generated by https://github.com/giantswarm/github, it will be merged automatically once all requirements are met


marians:
The problem here is that `Makefile` gets replaced completely. The existing Makefile should be renamed to match `Makefile.*.mk` and get imported by the main Makefile.

marians (approved):

